PR #8, opened 2017-03-08
Npm fix

### LINKED ISSUE #1: NPM package does not include build files
This package is exactly what I was looking for! Apologies that I'm about to file a handful of little issues.

If I `npm install` this package, I only get the typescript files. Not sure what your process is for releasing, for some reason the `build` directory is not included in the NPM package. Node complains of not finding `docx`

felipeochoa:
I think this resolves #1. If you're OK with it, I'll also rename the build directory to lib before merging.


dolanmiu:
make `prepublish` into `postinstall`

What prepublish was for transpiling the project, so people using JS can use it. `postinstall` perhaps makes more sense

felipeochoa:
Yeah though on second thought it's probably better to leave ts be so that the file histories are preserved. 
I think it makes sense to keep the `prepublish` script, since you want to transpile before publishing the package. That way users don't need to have typescript to install `docx`.

dolanmiu:
Seems like git preserves changes with the `--follow` argument:

http://thisbythem.com/blog/preserving-history-when-renaming-files-in-git/

Yes, you are right

just read the npm docs, it states that `devDependencies` are not installed by default (`typescript` is a dev dependency), so users cant `postinstall` to build the library

dolanmiu:
is there anything else which needs editing/adding. everything looks good to me

felipeochoa:
Hmm, interesting, did not know that was possible! Everything is good from my end, though I have not renamed anything. I can also give it a patch version bump so that you can push to NPM and see if the build dir is now included.

dolanmiu:
sure, if you bump it, ill publish tonight

if not, ill bump it tonight

PR #11, opened 2017-03-08
Fonts cleanup

felipeochoa:
Here's a smaller one for you:

* Moved run-fonts out of numbering and into docx/run
* linter warnings, linter warnings...
* Added `#font()` method to `Run`

The only really new thing is that now you can do:

```js
const run = new docx.TextRun().font('Times');
```


dolanmiu:
thanks! Why is Times New Roman 'Times'?

felipeochoa:
It's not. https://discussions.apple.com/thread/2169363?start=0&tstart=0
I picked 'Times' because it was shorter to type :smile:

dolanmiu:
wow I never knew this, thanks

PR #14, opened 2017-03-09
Clean up xml formatting (clearVariables)

felipeochoa:
Contrary to what I had written in the issue tracker, I did not have to implement the xml component classes as generics. As per your suggestion, I put the conversion methods on the classes themselves, named `#toXml`. I think they're quite short and understandable, but wait until you see what the formatter looks like now :smile:.

As with all the other changes, I have this running on my testing app, and I'm happy to say I can now re-use styles and numbering across document renders!


dolanmiu:
i am a little scared of the `clearVariables` removal, but `toXml` looks like more sense 😀

dolanmiu:
I am a little confused actually. I am using the popular package `xml` to convert the JSON to xml:

https://www.npmjs.com/package/xml

So why does .toXml() need to exist?

felipeochoa:
Yeah I think that's right. Submitting a PR with the changes, and addressing your comments about using `ret` as a variable name :blush:

PR #16, opened 2017-03-10
Cleanup

felipeochoa:
There are a few things in here; not controversial I hope!

* rename `toXml` to `prepForXml`
* removing the `console.log` statements: I think these were debugging aids from a long time ago. What I've seen libraries use in production is [`debug`](https://www.npmjs.com/package/debug), which lets you turn logging on and off by setting `DEBUG=docx:*` or `DEBUG=` in your environment
* set the compiler to use the `--strictNullChecks` option. There's a comment in felipeochoa/cleanup@3a6cf81 with more detail, but the gist it that it makes the compiler stricter


dolanmiu:
i added unused variable checking in tslint, to help with `clean up unused imports and a linter warning`

PR #17, opened 2017-03-10
More cleanup

felipeochoa:
This is a slightly bigger PR, only because I modified `XmlAttributeComponent`, which naturally touched a lot of files. The changes here are all cleanup, though:

* Not sure why, but `npm` and `install` were included as dependencies, which made fresh installs take forever. As they're not needed, I removed them
* With the latest refactoring, `lodash` is no longer needed (!!) so took that off as well. Hopefully this also fixes the "dependencies out of date" badge
* The `XmlAttributeComponent` adds significant type safety and makes subclasses a bit more DRY. More details in the comment for felipeochoa/docx@62911d6
* There were a couple of ergonomic improvements I made to the api:
    - The three packers (local, express, abstract) are now exported on the top-level `docx` object. Before, applications had to import them from `docx/build/export` which was a little clunky
    - I added a `document#createParagraph` method and also `paragraph#createTextRun` methods. Compare the before and after use of the api:
       ```js
       // before
       doc.addParagraph(new docx.Paragraph('Test heading1, bold and italicized')
                            .heading1());
       doc.addParagraph(new docx.Paragraph()
                        .addText(new docx.TextRun('Some monospaced content')
                                 .font('Monospace')));

       // after
       doc.createParagraph('Test heading1, bold and italicized')
           .heading1();
       doc.createParagraph()
           .createText('Some monospaced content')
           .font('Monospace');
       ```


dolanmiu:
Is it possible to use createParagraph() and new Paragrah?

Personally I like new Paragraph("text here"), but it's good to have options between the two

dolanmiu:
I'm guessing the next version bump will be a major bump 2.x.x as there's a lot of API changes 😄

dolanmiu:
In the old version of this project:

https://github.com/dolanmiu/Office-Clippy

It uses .createParagraph() but on larger documents it got so ugly 😂

But that's just me; maybe it would be suitable and preferred for other people.

But we should keep new Paragraph() as an option too

felipeochoa:
Yeah both options are still there! I just find `.createParagraph` lends itself better to chaining!

Re: API -- everything has been backwards compatible, so I don't think it's necessary to bump. Maybe 1.1.x? But I'd keep the 1 since we're still backwards compatible

dolanmiu:
Oh sorry yes, I thought you changed the name or TextRun class

felipeochoa:
Also, I don't want to push yet because it touches a lot of files, but I just cleaned up all the unused imports warnings from the linter

PR #21, opened 2017-03-11
Tables!

felipeochoa:
I think the table implementation is at a point where it can be merged. There are tests for everything, the api is nice to use, and most importantly, the output works correctly for the few cases I've tried. The implementation follows the spec, but there are possibly some quirks not captured by it (e.g., there was an issue where if tables don't specify column widths, LibreOffice won't render them as tables [already fixed]). Here's how you would use it:

```js
const table = doc.createTable(2, 3);
table.getCell(0, 0).createParagraph("This is the top-left");
table.getCell(0, 1).createParagraph().createTextRun("Heading 1 - italics").italic();
table.getCell(0, 2).createParagraph().createTextRun("Heading 2 - bold").bold();

table.getCell(1, 0).push(new Table(2, 3));  // Nested tables!!
table.getCell(1, 1).createParagraph()
    .createTextRun('Some content here, underlined')
    .underline('single');
table.getCell(1, 2).push(new Paragraph('This also works'));
```

### DIFF
--- a/ts/docx/document/index.ts
+++ b/ts/docx/document/index.ts
@@ -45,4 +45,11 @@ export class Document extends XmlComponent {
     public addTable(table: Table): void {
         this.body.push(table);
     }
+
+    public createTable(rows: number, cols: number): Table {
+        const table = new Table(rows, cols);
+        this.addTable(table);
+        return table;
+    }
+
 }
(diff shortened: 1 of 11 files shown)


dolanmiu:
look good to me! 

Good job

Shall we call it `.add()` rather than `.push()`

Because the rest of the API uses add, for example in `Paragraph`, it has an `addText` method

felipeochoa:
I renamed `cell.push()` to `cell.addContent()` -- I agree with moving everything to add. That way there's a clear parallel between `addXXX` and `createXXX`. I'll open up an issue to keep that in mind

PR #28, opened 2017-03-13
NPM fixes

felipeochoa:
This is a small fix to a couple of issues that I've encountered now that I'm using the library from NPM:

- Make sure to include the docx template in the NPM bundle
- Remove `app-root-path` and use `path.resolve` (full rationale in felipeochoa/docx@c7b7b58)

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
   },
   "files": [
     "ts",
-    "build"
+    "build",
+    "template"
   ],
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


dolanmiu:
Does path.resolve work properly? I remember there was a reason I used this library, but can't remember why😵

I am surprised that the library still works without putting the template folder in!

felipeochoa:
Yeah the `path.resolve` + `__dirname` trick works perfectly. I'd actually been running the library under a symlink to the development tree, so app-dir was picking up the uninstalled template folder. It was only when I tried npm install for real that it broke!

PR #41, opened 2017-09-15
Fix page/thematic breaks

jacwright:
These items are not paragraph properties, but part of the content of the paragraph.

I'm suggesting this fix, but I've not been able to build and test docx because of the error below, so please could you either direct me to how to get around the error or run the tests before merging?

```
ts/docx/xml-components/default-attributes.ts(20,32): error TS2536: Type 'string' cannot be used to index type 'AttributeMap<T>'.
```

### DIFF
--- a/ts/docx/paragraph/index.ts
+++ b/ts/docx/paragraph/index.ts
@@ -98,7 +98,7 @@ export class Paragraph extends XmlComponent {
     }
 
     public thematicBreak(): Paragraph {
-        this.root.push(new ThematicBreak());
+        this.properties.push(new ThematicBreak());
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


dolanmiu:
Yes you may have a point:
http://officeopenxml.com/WPtextSpecialContent-break.php
http://officeopenxml.com/WPparagraph.php

```xml
<w:r>
    <w:t>This is</w:t>
        <w:br/>
    <w:t xml:space="preserve"> a simple sentence.</w:t>
</w:r>
```

dolanmiu:
As for the test, I have to do some research, 

i'll merge this in soon, after some digging around, as it seems this pull request is not related to this broken test

dolanmiu:
Figured out the issue, because version `^2.4.1` of typescript broke it, as it used a newer version which had breaking changes.

I pinned it to `2.4.1` which fixed, merging your change now

Thanks!

dolanmiu:
Released `2.1.3` with this

PR #42, opened 2017-09-15
Fix tests

jacwright:
The tests weren't quite fixed here. Now that typescript is pinned and I'm able to run the tests, I fixed them.

Fixes tests for https://github.com/dolanmiu/docx/pull/41


dolanmiu:
Thanks, good stuff

PR #43, opened 2017-09-16
Making it easier to work with default styles

jacwright:
Adds the likely common styles, font and size, to be defined in defaults.


dolanmiu:
Nice work, this will be in `2.2.0`

PR #44, opened 2017-09-16
Allow indent to use other options than left and hanging

jacwright:
This is a breaking change. Existing code using indent will break when padding in the number for left.

If you don't like this breaking change I can submit another PR that just adds a 3rd parameter after `hanging` for the `firstLine` attribute. This attribute is the most common after `left` IMO, and the attribute I need. Book manuscripts indent the first line of each paragraph. I made this bigger change because I felt like 3 arguments, all optional, starts to get messy.

Please advise.

### DIFF
--- a/ts/docx/paragraph/indent.ts
+++ b/ts/docx/paragraph/indent.ts
@@ -3,22 +3,25 @@ import { XmlAttributeComponent, XmlComponent } from "../xml-components";
 interface IIndentAttributesProperties {
     left?: number;
     hanging?: number;
+    firstLine?: number;
+    start?: number;
+    end?: number;
 }
 
 class IndentAttributes extends XmlAttributeComponent<IIndentAttributesProperties> {
     protected xmlKeys = {
         left: "w:left",
         hanging: "w:hanging",
+        firstLine: "w:firstLine",
+        start: "w:start",
+        end: "w:end",
     };
 }
 
 export class Indent extends XmlComponent {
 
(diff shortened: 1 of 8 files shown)


dolanmiu:
I like this actually, it makes more sense

Does this change the API for the user? If so, then it has to be a `3.0.0` release thing, if not, then I could bump the version

jacwright:
It would have to be a 3.0.0 release. If someone is currently using `.indent(1234)` then it would no longer work for them.

dolanmiu:
I think it will be a little longer because I want to make it so you create a:

```
var document = new docx.File();
```

which would be breaking

I say this cos Microsoft Word has an odd way to handle images, by having references to other files. The `File` can contain the `Document`, and essentially be the manager for all the references.

`Document` is just an "xml file"

dolanmiu:
How about creating two methods, one old and one new, and when 3.0.0 comes out, remove the old one?

jacwright:
Done, this other PR will get me by until 3.0 is out. Thank you.

dolanmiu:
Thanks.

Will bump version now

jacwright:
This was the incompatible version! The other PR #45 is compatible with 2.x.

dolanmiu:
Oh god

dolanmiu:
`3.0.0` it is! 😂

dolanmiu:
`3.0.0` released

PR #45, opened 2017-09-16
Adding the ability to indent first line only in paragraphs

jacwright:
This is a backwards compatible fix while a better (but breaking) fix is here https://github.com/dolanmiu/docx/pull/44.


dolanmiu:
I made a project change to do with the tests, try get the latest version of the project

jacwright:
This is no longer needed with #44 merged. Closing.

PR #63, opened 2018-02-24
Added generic stream packer

felipeochoa:
This way `docx` doesn't have to explicitly create a new packer for every use case. (In fact, you could easily implement the express and fs packers on top of this one)

### DIFF
--- a/src/export/index.ts
+++ b/src/export/index.ts
@@ -1,3 +1,4 @@
 export * from "./packer/local";
 export * from "./packer/express";
 export * from "./packer/packer";
+export * from "./packer/stream";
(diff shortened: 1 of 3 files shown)


dolanmiu:
So stream packer is abstract? The end user wouldn't use this? Express and local packer would extend this?

felipeochoa:
No the idea is the consumer can do whatever they want with the stream. E.g., you can give the readable stream to a different web server, combine multiple streams into a tarball, upload the stream through Http or FTP, pass the stream to `unzip` (for testing), or anything else.

I'm not suggesting getting rid of the other packers, but was just pointing out that if the either one didn't exist an end user could easily have the same functionality in terms of this stream packer. (`stream.pipe(res)` or `stream.pipe(fs.createWriteStream(path))`)

dolanmiu:
Ah I see, makes sense!

felipeochoa:
Thanks!

PR #67, opened 2018-04-02
image scaling add rounding

ilmartyrk:
Found that sometimes after using image scaling documents where diplayed as broken by MS Word. It was down to image scaling units having  .5 values. After adding this fix no problems accured. Also exposed heading6 method.


dolanmiu:
Thanks!

dolanmiu:
Released in `3.4.0`

PR #70, opened 2018-05-06
H4 update

dolanmiu:
Kept PDF Packer

Project clean up to pass testing


src/file/index.ts
```diff
@@ -5,4 +5,3 @@ export * from "./numbering";
 export * from "./media";
 export * from "./drawing";
 export * from "./styles";
-export * from "./xml-components";
```
h4buli: If we don't export `XmlComponent` then there is no easy option for the users of the library to easy create new components. For example I would like to write:
```
class SomeOtherField extends XmlComponent {
}
```
that might have some logic that is specific for my usage, but not relevant for main library.

Or there is different way how can we can achieve this ?

h4buli:
Hi @dolanmiu, 
Just to give here an short overview for some of the changes:

**Numbering**
There is issue with Microsoft Word showing mixed list of bullet (unordered) and numbered (ordered) list -> if they are created in the not following the order:
 - First all abstract numberings
 - They all concrete numberings

**Working with external styles**
For some project and usage, it is convenient for the user to have e predefined template of styles. Adding a support to pass the styles (styles.xml of the extracted .docx document), provides lot of benefits for them.

**Images**
Add an option to pass image data as a Buffer (not to read from file).

**Tables**
Add an option to pass predefined column sizes. Extend table cell to support adding borders, vertical merge, horizontal span.

src/file/index.ts
dolanmiu: I see makes sense. I have never thought of that use case!

My initial thoughts were that it breaks encapculation, but your reasoning makes perfect sense

Will add back in

dolanmiu:
Could you write some demos regarding the external styles feature?

h4buli:
Hmm, I'm not sure how can I do that - since we need to read `styles.xml` from the file system (or load from somewhere else).  Let me write here the steps -> and you can point to me how can I make a demo.

In order to use external styles (based on some word template), you need to:
- Have a word document with some defined styles. We can override default word styles or we can add new one:
screen shot 2018-05-09 at 9 48 30 am


 - Save the word file, and then extract the `styles.xml` from the docx file (your template file with styles defined in word)
```
# unzip template.docx -d templateFolder
```
 - Read `templateFolder/word/styles.xml` and pass when creating new `Document`
```
import * as fs from 'fs';
import * as docx from 'docx';

const styles = fs.readFileSync('word/styles.xml', 'utf-8');
const document: docx.Document = new docx.Document({
  title: 'Title',
  externalStyles: styles
});

let p = new docx.Paragraph('This is a paragraph that has styles from the template');
p.style('MyFancyStyle');

var exporter = new docx.LocalPacker(document);
exporter.pack('My Document');
```

h4buli:
Great 👍

PR #75, opened 2018-05-17
paragraph: add support for 'page break before' format property

h4buli:
This PR adds support to add `Page break before` method available on the Paragraph. 
This option (available in word) will make sure that the paragraph will start on a new page (if it's not already on a new page).

Word option:
screen shot 2018-05-17 at 2 08 28 pm

More info:
It's different solution then adding a Page break element manually, since we can have use cases:
 - it's the first paragraph (makes no sense to have a blank page)
 - there is already page break element added

### DIFF
--- a/demo/demo15.js
+++ b/demo/demo15.js
@@ -1,0 +1,14 @@
+const docx = require('../build');
+
+var doc = new docx.Document();
+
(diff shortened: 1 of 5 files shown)


dolanmiu:
Thanks!

I have added a demo as well

Merged

PR #86, opened 2018-06-22
Feat/multiple headers

dolanmiu:


formatically:
I see that the header along with "different first page header" functionality is being simplified.  Anything I can do to help?

dolanmiu:
The people over at H4 @h4buli have devised a system where you can have multiple headers, footers and sections

Which means not only you can do `Different First Page Header`, but an infinite amount of headers and footers, which is good!

What this means though is that there is a slight change to the API for different first page header

formatically:
Sweet, that's awesome!

PR #99, opened 2018-07-24
Add RTL support for text run

amitm02:
In addition to bidi support for paragraphs, in RTL languages, each text run should contain an RTL tag.
http://www.datypic.com/sc/ooxml/e-w_rtl-1.html


dolanmiu:
can u fix the build issues, then I will merge

Issue #116, opened 2018-08-07
Error adding a image to a table


AladMocu:
i got `docx.Media.addImage is not a function` everytime I try to add a image to a table or the media

```
docx.Media.addImage is not a function
    at Interface.rl.on.on `(D:\testdocs\main.js:32:30)`
    at Interface.emit (events.js:187:15)
```

im following the example 

```js
const docx = require('docx');

var doc = new docx.Document();

const table = doc.createTable(4, 4);
table.getCell(2, 2).addContent(new docx.Paragraph('Hello'));

const image = docx.Media.addImage(doc, "./demo/images/image1.jpeg");
table.getCell(1, 1).addContent(image);

var exporter = new docx.LocalPacker(doc);
exporter.pack('My Document');

console.log('Document created successfully at project root!');
```

dolanmiu:
This isnt released yet, will be available in 4.0.0

AladMocu:
Thanks for fast response, meanwhile, is there any way to use the ```doc.createImage``` whitout the line break after the image?

dolanmiu:
That is a good question, and the answer is no

This is because an "Image" in `docx` is actually a paragraph, which explains the line break

I will address this in `4.0.0`

Thanks for pointing this out

PR #118, opened 2018-08-09
Add image to run

dolanmiu:
Ability to add inline image to run

Closes https://github.com/dolanmiu/docx/issues/116


AladMocu:
Sorry,How do you use that ?

dolanmiu:
I will explain in a guide later at http://docx.js.org

PR #126, opened 2018-08-14
Added optional drawingOptions parameter to addImage and addImageFromBuffer methods

sergiofm:
Added optional drawingOptions parameter to addImage and addImageFromBuffer so this way it can be passed to the ImageParagraph constructor, giving the ability to configure how the image will be inserted in the document for this methods.


dolanmiu:
Hey there,
`addImage` will be deprecated soon because it uses `fs`.

I am currently making a PR which will make it possible for this library to be used in the browser (Angular/React/HTML) etc!!

This means I need to sadly refactor it in such a way `fs` references aren't in `docx` (e.g. remove `addImage `)

So I need to put your PR on hold for a little bit

sergiofm:
Hi @dolanmiu! 
No problems with holding the PR!
My use case is pretty complex, so I believe it will help to improve a lot docx.
If you're interested in the use case, we developed a online editor for academic works, and we need to export in word format. Here's our [site](https://www.mettzer.com). You can register to see the editor.
I'm studying the sources and I will be really happy to help you in the development of docx.
Is there any channel where we can talk about it?

PR #128, opened 2018-08-19
Feat/paragraph borders

nschechter:
Hey there, I really needed this for a personal project - maybe y'all think if it's good enough it can be merged in. Paragraphs need to be explicitly given a border before users are able to create borders for it.

Thanks again for the great work on this library. I appreciate all the contributors


dolanmiu:
Nice good stuff!

I want to merge this first:

https://github.com/dolanmiu/docx/pull/124

So this may have to wait a few days if thats ok

PR #143, opened 2018-09-05
use .dotx files as an initial template for a new (document) file

amitm02:
I've started to implement a way to use an existing template document for a new (document) file.
Please see demo27.

I've concentrated on it being able to init the headers and footers of a new document from a template. But i think it be can easily extended in import other parts (e.g styles/numbering etc').

I think it can be a great uplift to the library. Currently the library does an amazing job at generating .docx files (JS -> docx). This could be the beginning of building the other direction of the bride:  (docx -> JS)

**IMPORTANT: although the demos seems to work. But the update will fail test/linting/pretty runs. I hope i can have help with that.**

How it works:

- /src/importDocx/importDocx.ts, includes a new class that accept a ".dotx" Buffer, parse it, and returns a simpler version of a template "File" class: 
```
interface TemplateDocument {
    currentRelationshipId : number;
    headers : {type : HeaderReferenceType, header : HeaderWrapper}[],
    footers : {type : FooterReferenceType, footer : FooterWrapper}[],
}
```

- The parsing is "narrow" and not "wide", in the sense i've only extracted headers and footers. Making the parsing "wide" to the point in generate a full "File" class (e.g styles) can be an future feature.

- The parsing is "shallow" and not "deep", in the sense that the content under an extracted "HeaderWrapper", is stored as the base "XmlComponents" and not as the extended version (e.g "Paragraph", "Text"). Making the parsing "deep" can be an future feature.

- What the above means is that it is possible the modify the headers/footers on the top level (e.g add a new paragraph), but modifying an inner XML in them (e.g an exciting paragraph) is not yet possible.

-  The "File" constructor, can accept this "TemplateDocument " object as a property in "options" parameter, and it will build a new document have having those headers/footers (including media that they may have).

Some other changes that was needed to make the library more align with possible template inputs:
- A section can have multiple headers/footers (possible types: default, first, even).
- Each header/footer has its own "media" (images).

One last note: Parsing an exiting .docx files they way is a good method to revel hidden bugs. For example, while testing the new code' I've found that the new File that was generated from a template was corrupted. The reason was that the line in prepForXml() 
`".filter((comp) => comp)" // Exclude null, undefined, and empty strings`
was filtering the value 0 in addition to null and undefined. That resulted in a invalid xml for docx.


dolanmiu:
This seems to break ./demo/demo16.ts

amitm02:
looking into it...

amitm02:
a fix for demo 16 was pushed.

dolanmiu:
I will fix the linting errors tonight :)

I am a bit concerned about the breaking unit tests though:

```
+ expected - actual

       {
      -  "w:tblPr": [
      -    ""
      -  ]
      +  "w:tblPr": []
```

Why is there something in the array, doesn't feel right

amitm02:
It is strange. I have not touch any of the tables code. I’ll try to find time to look at it tomorrow.

dolanmiu:
Read my comment on the Media object

amitm02:
> Read my comment on the Media object
where is the comment?

dolanmiu:
>I am not sure about this
Ideally there is only one Media singleton in the File class and not in the header and footers
Imagine if you add the same image in header/footer/document? it will be duplicated 3 times

amitm02:
> Ideally there is only one Media singleton in the File class.

The problem is that currently, "IMedia" contains a referenceID property, and ideally it shouldn't.
In OOXML, The document, headers and foots has their own references.
The Document body can assign one referenceID to an image (let say "rId7").
And an Header can assign a different referenceID to the exact same image (let say "rId1").

Removing the referenceID from IMedia, would require an even greater code change than i already done, and I felt i was pushing as is for a single/first merge.

Indeed, the price to pay for the current solution is that in the rare(?) case that the exact same image is used in the body and header: it is duplicated. It seemed reasonable to me as a first step. But ultimately i think i right approach is to get referenceID out from IMedia.

dolanmiu:
You have a valid point here,

Ok, we will keep the Media in header/footer

> "IMedia" contains a referenceID property, and ideally it shouldn't.

The reason it has a `referenceID` is because in the scenario you want to exact same image twice, you would want to re-use the reference right? 

Check out this demo, image5 is used twice

https://github.com/dolanmiu/docx/blob/master/demo/demo23.ts#L21

At the time, it seemed reasonable, I am open to other suggestions!

dolanmiu:
> The header can use rId1 for the image and the document can use rId7 for the same image.

Oh i see, same image, different ref, thats interesting. May need a lot of work, 

need to do this in future, after this PR

amitm02:
Apologizes for not being more active on this PR. It is holiday now in Israel not finding time to work on this project is a bit hard..

dolanmiu:
It's fine! Enjoy your holiday, no rush

amitm02:
There are some testing errors i don't manage to fix regarding to "fastXmlParser", parseOptions: "textNodeName".
https://www.npmjs.com/package/fast-xml-parser.

Do you have any experience with that library?

dolanmiu:
Does this still work for you?

I can see an error:

image

dolanmiu:
I will look into these failing tests after

dolanmiu:
Good news is, I fixed the tests, bad news is, it needs `.filter((comp) => comp !== null);` to work

Finding a way to fix

amitm02:
amazing!!!
much respect!

dolanmiu:
Question though:

```
  headers: [{headerId: header.Header.ReferenceId, headerType: HeaderReferenceType.FIRST}],
    footerId: footer.Footer.ReferenceId,	    footers: [{footerId: footer.Footer.ReferenceId, footerType: FooterReferenceType.FIRST}],
```

Why in demo16.ts, it takes in an array of headers and footers now? Is it one per each page or something?

amitm02:
From: lhttp://officeopenxml.com/WPSectionHeaderReference.php


For each section there can be up to three types of headers: first page header, odd page header, and even page header. The type is specified via the type attribute.

So if you only want just one header, you can use an array with one header element (type default). But you can also have a different header just for the first page, ln that case, you’ll add another header element in the array with type “first”

dolanmiu:
> three types of headers: first page header, odd page header, and even page header

Ok, can I change the API to something like:
```
headers: {
  first: header,
  odd: header2,
  even: header3
}
```

I think it's a bit nicer than arrays and ReferenceId etc

amitm02:
It is possible. I guess it all depends on how you want to represent something liked this:

<w:sectPr>
. . .
<w:headerReference r:id="rId3" w:type="first"/>
<w:headerReference r:id="rId5" w:type="default"/>
<w:headerReference r:id="rId2" w:type="even"/>
. . .
</w:sectPr>

Both an array or a map can work. An Array is more in line with the XML structure, a Map can be nicer wrapper for the users. 

Just note that the possible types are:
Default, first and even. Depending on which are defined, different  behavior is possible. It is actually surprisingly well documented in http://officeopenxml.com/WPSectionHeaderReference.php

dolanmiu:
> Map can be nicer wrapper for the users.

My vision for this library is to give users a easy no-brainer API 😄 

I have changed the API a bit if you don't mind. 

Now there is `File` and `TemplatedFile`

I felt that `File` was getting too bloated, long and doing too many things. The constructor took me too long to read. 

So I separated the Template stuff out into a new File type. Separation of Concerns 🎉

dolanmiu:
And btw, why is it called `ImportDocx`, should it not be called `ImportDotx`? lol

```
const filePath = "./demo/dotx/template.dotx";
```

dolanmiu:
Images in tables and footers seem to be bad now

image

dolanmiu:
I'll fix merge conflicts tomorrow

amitm02:
Amazing job man! You are quick. 

“And btw, why is it called ImportDocx, should it not be called ImportDotx? “

We’ll it should suppose to work with .docx files too. To my understanding, the only difference between dotx and docx is that you can not save to update (override) the content of a dotx file.

amitm02:
“Now there is File and TemplatedFile”

Perhaps it is a good idea for now, but looking forward, I think that looking it at as “templated file” is a wrong approach. 

As I’ve mentioned before, dotx (templates) are almost the same as regular docx files. 

What this pull request really allows, is the ability load and update an a existing word document, instead of only creating a new one. 

Granted, it is now limited to loading headers and footers, but with some more work, it can include parsing of the document body too. See my original post in this pull request.

dolanmiu:
> As I’ve mentioned before, dotx (templates) are almost the same as regular docx files.

Yes thats why `File` and `TemplatedFile` inherit `BaseFile`, they are internally the same type

What do you think?

I can think of another way to reduce the complexity of `File`

my main concern is that sooner or later `File` will be unreadable 500+ line class

amitm02:
I think that as the owner of the library, you’ll have to make a decision down the road. Is it a library primarily for writing new document, or is that a library that can read and write documents. If it is the former, having a separate “templated file” is the right approach.  If it is the latter, I think the File class should have the ability to initialized with an existing document (hence constructor). 

I’ll support you on each path you’ll choose :)

dolanmiu:
Yes I would like it to read and write files one day, would be cool

After some thought, I think you may be right

I will revert back 😄 

Thank you for your support

dolanmiu:
Wow, thanks so much for discovering:

```
".filter((comp) => comp)" // Exclude null, undefined, and empty strings
```

I think it was actually a **hack** made by someone to remove unwanted `XMLComponents` by setting them as `""`

I was always wondering why there could be a string value in there, thought it mightve been a config or flag, but nope, a hack 😱

dolanmiu:
Fixed tests and it is all working!!! 🎉

amitm02:
I have an error on file.d.ts:
`import { IFileProperties } from "./file-properties";`

file-properties.d.ts does not exists..

dolanmiu:
I'll look into this today

dolanmiu:
Fixed the `IFileProperties` issue 🎉

amitm02:
Nice!, i've missed that.
Is there anything missing that is holding us from doing a merge?

amitm02:
demo29 does not work for some reason... it says file is corrupted..
commit #0a8feca works. something after it broke it...

dolanmiu:
nope, it changed the API to a nicer one

Will work on the image issue once i get some free time

dolanmiu:
Did some research, its the way the ids are done. Now it's always `rId0`, before it increments.

I have a feeling this is because now there are multiple `Media` objects

```xml
<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
    <Relationship Id="rId2" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/image" Target="media/q5iqxb2wea5ohdhv7elt.png"></Relationship>
</Relationships>
```

```xml
<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
    <Relationship Id="rId0" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/image" Target="media/f9xcpma1qrpvg6hx1gzbuj.png"></Relationship>
</Relationships>

```

amitm02:
the last commit (97f76fb) cause a broken docx for some cases. e.g change template at demo30 to the attached file

[e.dotx.zip](https://github.com/dolanmiu/docx/files/2482134/e.dotx.zip)

dolanmiu:
ill have a look tonight

dolanmiu:
Fixed, because i forgot to add `break` after case, silly me 😂

dolanmiu:
Very nice!

amitm02:
> Why is fromXmlString method deleted?
> 
> It could be required by someone else
> 
> Why are the tests commented out?
> 
> We have to un-comment it

fromXmlString was not used anywhere in the code.
Even before my change to xml-js, the parsing of xml string to XmlComponent was done via convertToXmlComponent(). It offers no additional value beyond what convertToXmlComponent already offers. I think it was there for historical reasons..

amitm02:
Hi @dolanmiu, anything stopping this pull request from reaching its (happy) ending? :)

dolanmiu:
Most of it is merged here :smile: :

https://github.com/dolanmiu/docx/pull/179

It's just the fromXmlString I have a problem with

amitm02:
As i said, its not used anywhere. Don't take my word on it (😄), you can search the code base and see that the string "fromXmlString" does no appear anywhere.

dolanmiu:
But it could be used externally right? 

What about this use-case:

Someone unzipped his `.docx` file, and parses `document.xml` into a `string` and then would like to create an `XmlComponent` out of it?

amitm02:
> But it could be used externally right?
> 
> What about this use-case:
> 
> Someone unzipped his `.docx` file, and parses `document.xml` into a `string` and then would like to create an `XmlComponent` out of it?

Then he will use "convertToXmlComponent()". This function accept and XML string and return an XmlComponent.

dolanmiu:
> Then he will use "convertToXmlComponent()". This function accept and XML string and return an XmlComponent.

This function does not accept string, see screenshot below, it accepts `Element`:

image

dolanmiu:
Merging!

amitm02:
🎉

h4buli:
Hi all, 

I'm little late to the party - sorry been busy. But I've seen that it's merged and it looks perfect.

Just to add some input from my side:
 - `fromXmlString` is static method on `ImportedXmlComponent` that is used to create new instance of the class using String. 
`const imported = docx.ImportedXmlComponent.fromXmlString(xmlFileAsString);`
 - internally  `fromXmlString` is using `convertToXmlComponent` to create from XmlElement -> XmlComponent

xmlString ---> **fromXmlString**() ---> xmlElement ---> **convertToXmlComponent**(xmlElement) ---> XmlComponent 

We are using `fromXmlString()` to read and parse full page watermark (created previously with Microsoft Word) and to embed it to the document (in Header) :-)

PR #149, opened 2018-09-13
Feat/grid span

dolanmiu:


felipeochoa:
Not sure if you've finalized this API, but one suggestion (to avoid the ugly Table reference in the cell) would be to instead create a `table.mergeCells(x0, y0, x1, y1)` method to merge a block of cells at the Table level. I think merging cells like that is more intuitive/ergonomic than having to set horizontal spans and such. If merging in both directions simultaneously is difficult, then maybe just `row.mergeCells(from: number, to: number)`

dolanmiu:
That's a nice suggestion

You're right, it feels much more natural as well

I will do this

Good spot!

PR #153, opened 2018-09-17
Fix 109

felipeochoa:
This PR (merge after #152) allows defining styles entirely in Typescript

### DIFF
--- a/demo/demo27.ts
+++ b/demo/demo27.ts
@@ -1,0 +1,34 @@
+import * as fs from "fs";
+import { Document, Packer } from "../build";
+
+const doc = new Document();
+const myStyles = doc.Styles;
+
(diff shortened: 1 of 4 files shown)


dolanmiu:
Whats the advantage of doing:

```
myStyles = doc.Styles

myStyles.createParagraphStyles....
```

Rather than just do:

```
doc.Styles.createParagraphStyles...
```
Surely this is better: https://github.com/dolanmiu/docx/blob/master/demo/demo2.ts

I can see some potential problems because there is a setter for styles now, which means someone can do:

```
doc.Styles = {} 
```
and break something?

felipeochoa:
I use it to define the styles once and then use them for multiple different documents. The latter error would be caught by Typescript

PR #161, opened 2018-09-20
Improve fluency (dsl like)

bre7:
`addRunProperty` / `addParagraphProperty` should also return `this`


dolanmiu:
fix the linting/style errors then ill merge

will release today

PR #175, opened 2018-10-19
Created classes and methods needed to create Sequential Identifiers

sergiofm:


dolanmiu:
What does this feature do?

I get the following, which seems like its just normal text?

image

dolanmiu:
Oh I understand now!

PR #179, opened 2018-10-22
Feat/import dotx

dolanmiu:


dolanmiu:
I am going to merge this PR for now as it is getting very big, and is very long due.

I have decided to use a single Media object now because Media is actually a concept I made up, so it does not relate to Microsoft Word at all. It is simply syntactic sugar to help users.

Great work on this feature!

PR #183, opened 2018-10-25
method to remove pagebreaks

sergiofm:


dolanmiu:
For what use-case would this be useful?

sergiofm:
Sometimes, in my logic, I need to remove page breaks from pre-created paragraph.

dolanmiu:
Is it possible to not create the page breaks in the first place?

I just have a weird feeling about this method

What could be next?

```js
paragraph.removeAllText()
```

```js
paragraph.removeAllImages();
```

😂

dolanmiu:
thanks

PR #197, opened 2018-11-09
Bugfix - Correct default section options

sergiofm:
The page number type attribute of the sections was always been created, leading Word to always reset page numbers to zero in new sections.
The page number type DECIMAL is already default in Word, there is no need to force this to be the default option in the default section, like it was.
Demo 16 was updated to show how page number types format and start work.
Improvements in headers and footers docs

### DIFF
--- a/demo/demo16.ts
+++ b/demo/demo16.ts
@@ -1,7 +1,7 @@
 // Multiple sections and headers
 // Import from 'docx' rather than '../build' if you install from npm
 import * as fs from "fs";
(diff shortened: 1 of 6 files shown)


sergiofm:
Seems like there's some problem in `npm run style` at travis ci.
I ran it locally and there's no problem here.

dolanmiu:
Yes I have noticed the weird `npm run style` thing too

Locally it works fine for me, I have now fixed it though

dolanmiu:
I will merge this as from reading CI logs, all tests pass, and only the `npm run style` fails (which is now fixed)

PR #205, opened 2018-11-13
Refactor in Style

sergiofm:
Refactor in style to separate classes in their specific files.
Tests improvements.


dolanmiu:
can you do an `npm install` and format again, I fixed that style issue

PR #214, opened 2018-12-03
Update xml-component.ts

### LINKED ISSUE #213: Breaks include undesired spaces
`paragraph.createTextRun().break();` is adding a bunch of spaces on the line of the break.

_I think_ it might be because a linebreak is added to the `document.xml` like this:

```xml
                <w:br>
                </w:br>
```

and should be like this:

```xml
                <w:br></w:br>
```

or

```xml
                <w:br />
```

jacwright:
Empty XML nodes are not closed and can have spaces in them which adds unwanted spaces in Word documents when indenting. This fixes #213.


jacwright:
Closed because #213 was incorrect.

PR #229, opened 2019-01-02
Fix getting started docs for v4 packer.

microbit-matt-hillsdon:
I think it's better if these docs work for v4 as that's what new users will get via npm install.


dolanmiu:
Thanks!!

PR #241, opened 2019-01-10
Added compiler options and set target for ES5 transpile

alperunal:
ES6 target does not supported by IE 11

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es6",
+        "target": "es5",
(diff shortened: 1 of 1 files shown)


jankeesvw:
I tried this branch, but run into this error. Is there a work around to get this working in IE11?

```js
    Cannot find module 'docx' from 'exportToWord.js'

    > 1 | import { Document, Packer, Paragraph, TextRun } from "docx"

dolanmiu:
Closing as `docx` does not plan to support IE11 in future as Microsoft has dropped support for IE11

Issue #232, opened 2019-01-04
Bugs found & some feature requests


kkoomen:
Hi there. 

First of all: good job on v4.4 lately compared to version 3. We recently updated to v4 within our company and the v4.4 has a lot of new things and improvements. Unfortunately we came across some issues we really wonder why they're even documented because they do not seem to work.

We use this module _heavily_ within our company for one of our clients so we are getting the most out of it and we had to debug a lot in the core back then with v3 (also with v4 already) and within docx its XML files to see what was wrong. Below you'll find our feedback.

## Some background information

- We use this module in NodeJS.
- We personally test using the latest version of LibreOffice on OS X. We also test on Microsoft Word 2007+ if a functionality is complete and finished for LibreOffice.

### Numbered lists
They seem to render almost correctly in the XML but it needs.. _something_. We're not sure what, but the problem is that the numbers do not show. However, bullets render correctly.

If I am using the jsfiddle provided in the README and adjust the code to be the code from [this](https://runkit.com/dolanmiu/docx-demo3) demo and use v4.4.1 it gives me this when I open it in libreoffice:

<img width="614" alt="screenshot 2019-01-04 at 23 17 39" src="https://user-images.githubusercontent.com/10693490/50713951-2803c880-1077-11e9-8bf8-213c4864a8a2.png">

So the demo isn't working already, hence the fact it doesn't work for us. If this could be fixed that would be great.

### Table of Contents

This was a huge add-on that we loved to see, unfortunately it doesn't work as well. Even when copying demos into the jsfiddle again, it still doesn't work. My colleague did the hard debugging for this one and mentioned that the XML doesn't generate the hyperlinking properly. Also, a related problem is that my document is not even recognizing the `.heading1()` as an actual heading. When I am looking in my overview of my document it says it has 0 headings, while it has multiple headings defined. This is probably the _main_ reason for the Table of Contents not being able to work.

### Lists
A thing we would like to adjust ourselves is the option _"Don't add space between paragraphs of the same style"_ in a _"List Paragraph"_. I am seeing that this issue #222 is already marked as a _"Feature request"_ so I hope this will be implemented as well. 

I have found the actual XML issue that causes this. The list items have in the xml `xml:space="preserve"` and this attribute will be removed when _"Don't add space between paragraphs of the same style"_ is checked. If you guys are able to implement this for _"ListParagraphs"_ that would be great!

[...]

PR #248, opened 2019-01-15
Add OutlineLevel Paragraph property

filippomuscolino:
Hi! 
We were working on TOC and we've seen that it doesn't work in LibreOffice (as @kkoomen said in [#232](https://github.com/dolanmiu/docx/issues/232#issuecomment-451654405)).
We find out that a solution could be setting the property "outlineLvl" for paragraph.

Something like:
```
const toc: TableOfContents = new TableOfContents('Index', {
   hyperlink: true,
   headingStyleRange: '1-5'
});
this.doc.addTableOfContents(toc);

const paragraph: Paragraph = this.doc.createParagraph('Chapter 1').heading1();
paragraph.outlineLevel('0');
```

What do you think about?

### DIFF
--- a/demo/demo39.ts
+++ b/demo/demo39.ts
@@ -1,0 +1,33 @@
+// Scaling images
+// Import from 'docx' rather than '../build' if you install from npm
+import * as fs from "fs";
+import { Document, Packer, PageNumberFormat, TextRun } from "../build";
+
+const doc = new Document(
+    {},
+    {
+        pageNumberStart: 1,
+        pageNumberFormatType: PageNumberFormat.DECIMAL,
+    },
+);
+
+doc.Header.createParagraph("Foo Bar corp. ")
+    .addRun(new TextRun("Page Number ").pageNumber())
+    .addRun(new TextRun(" to ").numberOfTotalPages());
+
+doc.Footer.createParagraph("Foo Bar corp. ")
+    .center()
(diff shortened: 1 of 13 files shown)


dolanmiu:
Thanks for this!

Do you think outline level should be done by default then?

filippomuscolino:
> Edit: So for it to work on Libre Office, people have to explicitly add outline also?

Yes, cause it seems that LibreOffice generates the TOC based only on outline levels (we read it [here](https://ask.libreoffice.org/en/question/91168/empty-table-of-contents-even-thought-i-set-all-the-headings-correctly/))

> 
> should heading1() add outline as well?

Yes, maybe could be better to add outline level property by default for every heading

dolanmiu:
Makes sense

> Yes, maybe could be better to add outline level property by default for every heading


I like this approach, makes it easy and clean to use

So heading1 = outlineLevel 0
heading2 = outlineLevel 1
heading3 = outlineLevel 2? etc

filippomuscolino:
Exactly! 
The only problem that just came in my mind it's that if in the TableOfContentOptions you put something like headingStyleRange: '3-5', LibreOffice will keep using OutlineLevel, so all the headings will be in TOC.

A better solution would be setting the OutlineLevel according to what is specified in headingStyleRange property, but I'm not sure if it's possible

dolanmiu:
Ok, I'll merge this in now as it is useful

PR #265, opened 2019-02-08
dont run coverage if publishing

joefitter:
istanbul-instrumenter-loader is running on _every_ build. This means coverage paths are ending up in the published built module - this causes CSP to fail due to strings such as `"C:\\Users\\Dolan\\Documents\\docx\\src\\index.ts"` in the built module.


dolanmiu:
Thanks!

Good spot

Issue #236, opened 2019-01-09
Merging cells has unexpected bahavior


ricardotik:
Hi 
I appreciate this awesome library to generate docx in JS/TS easily.
I am working on table cell manipulation involving cell merging.
I found out the below unexpected behavior when I try to merge `col 0` to `col 4` for a 6-column table. 
There is not any problem for merging between `col 0` to `col 2`. 
The last col will be omitted when merging `col 0` from `col 3`  to `col 4`.

Below is the code listing that demonstrate the problem.
```
const table = doc.createTable(13, 8);
row = 0;
applyAllBorder(table.getCell(row, 0).addContent(new docx.Paragraph("0,0")));
applyAllBorder(table.getCell(row, 1).addContent(new docx.Paragraph("0,1")));
applyAllBorder(table.getCell(row, 3).addContent(new docx.Paragraph("0,3")));
applyAllBorder(table.getCell(row, 4).addContent(new docx.Paragraph("0,4")));
table.getRow(row).mergeCells(4, 5);
table.getRow(row).mergeCells(1, 2);
row = 1;
applyAllBorder(table.getCell(row, 0).addContent(new docx.Paragraph("1,0")));
applyAllBorder(table.getCell(row, 2).addContent(new docx.Paragraph("1,2")));
applyAllBorder(table.getCell(row, 4).addContent(new docx.Paragraph("1,4")));
table.getRow(row).mergeCells(4, 5);
table.getRow(row).mergeCells(2, 3);
table.getRow(row).mergeCells(0, 1);

row = 2;
applyAllBorder(table.getCell(row, 0).addContent(new docx.Paragraph("2,0")));
applyAllBorder(table.getCell(row, 1).addContent(new docx.Paragraph("2,1")));
applyAllBorder(table.getCell(row, 2).addContent(new docx.Paragraph("2,2")));
applyAllBorder(table.getCell(row, 3).addContent(new docx.Paragraph("2,3")));
applyAllBorder(table.getCell(row, 4).addContent(new docx.Paragraph("2,4")));
table.getRow(row).mergeCells(4, 5);
table.getRow(row).mergeCells(1, 2);
row = 3;
applyAllBorder(table.getCell(row, 0).addContent(new docx.Paragraph("3,0")));
applyAllBorder(table.getCell(row, 1).addContent(new docx.Paragraph("3,1")));
applyAllBorder(table.getCell(row, 2).addContent(new docx.Paragraph("3,2")));
[...]

PR #273, opened 2019-02-26
Add higher width for GridCol to fix merging cells

dolanmiu:
Fixes https://github.com/dolanmiu/docx/issues/236


ricardotik:
appreciated!

PR #275, opened 2019-03-05
Table pagination: add cantSplit and tblHeader row properties

### LINKED ISSUE #263: Table pagination: don't break row and repeat header on each page
Hi!
It would be great if you could add these properties for TableRow (see more [here](http://officeopenxml.com/WPtableRowProperties.php)):

`<w:cantSplit w:val="true" /> `
If present, it prevents the contents of the row from breaking across multiple pages by moving the start of the row to the start of a new page

`<w:tblHeader w:val="true" />`
Specifies that the current row should be repeated at the top each new page on which the table is displayed

Thanks!
Filippo

filippomuscolino:
Added two properties for TableRow: 

- **cantSplit** (prevents the contents of the row from breaking across multiple pages)
- **tblHeader** (the current row should be repeated at the top each new page on which the table is displayed)

Example:
```
table.getRow(0).setCantSplit();
table.getRow(0).setTableHeader();
```

Closes #263

### DIFF
--- a/src/file/table/table-row/table-row-properties.spec.ts
+++ b/src/file/table/table-row/table-row-properties.spec.ts
@@ -1,0 +1,31 @@
+import { expect } from "chai";
+import { Formatter } from "export/formatter";
+import { TableRowProperties } from "./table-row-properties";
+
+describe("TableRowProperties", () => {
+    describe("#constructor", () => {
+        it("creates an initially empty property object", () => {
+            const rowProperties = new TableRowProperties();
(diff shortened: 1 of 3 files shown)


dolanmiu:
Thanks!

Again, coverage is lacking, but I will accept as I am planning to revamp tables!

filippomuscolino:
Thanks! I have some PRs left that I'll publish this afternoon, I'll try to improve the coverage then!

PR #277, opened 2019-03-05
Select styles that will go into TOC

filippomuscolino:
In order to manually add a paragraph into the TOC you can specify outline level in paragraph properties (as already discussed on #248).

It can be useful to specify this property also into the paragraph style, so that all paragraphs using that style will go into the TOC.

_Example_

`this.doc.Styles.createParagraphStyle('myStyleThatGoesInTOC').outlineLevel('1');`


dolanmiu:
Thanks!

PR #279, opened 2019-03-05
Fix: settings are ignored by Word without relationship and content type

filippomuscolino:
After adding some settings to _settings.xml_ file, we realized that Word was ignoring those settings. After some research it came out that Word needs also that the settings relationship and content type are specified.

More precisely: in _document.xml.rels_ the relationship missing was:
`<Relationship Id="rId4" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/settings" Target="settings.xml"></Relationship>`

While in _[Content_Types].xml_ the content type missing was:
`<Override ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.settings+xml" PartName="/word/settings.xml"></Override>`

### DIFF
--- a/src/file/content-types/content-types.spec.ts
+++ b/src/file/content-types/content-types.spec.ts
@@ -79,6 +79,26 @@ describe("ContentTypes", () => {
                     },
                 ],
             });
+            expect(tree["Types"][13]).to.deep.equal({
+                Override: [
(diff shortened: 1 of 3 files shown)


dolanmiu:
Perfect!

Good work

PR #280, opened 2019-03-05
Add setting for not justify lines ending in soft line break

filippomuscolino:
Add a setting to do not justify the contents of incomplete lines, which end in a soft line break, when the parent paragraph is fully justified. Works in LibreOffice and Word.

Steps to reproduce this In Word 2013:
- Click the File tab, Options, Advanced.
- Go to absolute bottom of the Advanced options, and click to expand Layout Options.
- Tick “Don’t expand character spaces on a line that ends with SHIFT-RETURN“

**Before**
before

**After**
after

_Example_

`this.doc.Settings.addCompatibility().doNotExpandShiftReturn()`


dolanmiu:
Thanks!

PR #282, opened 2019-03-06
Fix: table width in percentage should include '%'

filippomuscolino:
Hi! 
Trying the last version I found out that table with width in percentage didn't work anymore.
It seems that from 2011 version of the standard OOXML when type="pct" you have to include % (more info [here](http://officeopenxml.com/WPtableWidth.php)).
I tried with both Office 2019 and LibreOffice last version.

I opened a PR instead of an issue. If you prefer you can fix it in other ways and just discard this.
Thanks


dolanmiu:
Good spot

Can you fix the tests?

PR #299, opened 2019-04-04
Fix: duplicated generation of last section properties

### LINKED ISSUE #259: Header/Footer duplicated
Hi,

I'm using last version of docx and I noticed that the Header/Footer is duplicated (at least in LibreOffice, did not try in Office).
It seems that the XML has <w:sectPr> duplicated, with the same content.
(To reproduce this, try for instance demo14)

Could you please check it? Thanks!

Filippo

filippomuscolino:
When generating XML for body, section properties for last section are duplicated (see #259).
It seems that you have to empty sections array after adding last section to body (don't know why the formatter is passing two times there).

This solves the problems of 

- duplicated header/footer of last section in LibreOffice - Fixes #259 
- wrong header/footer of last section in Word

### DIFF
--- a/src/file/document/body/body.ts
+++ b/src/file/document/body/body.ts
@@ -37,7 +37,7 @@ export class Body extends XmlComponent {
     }
     public prepForXml(): IXmlableObject | undefined {
         if (this.sections.length === 1) {
-            this.root.push(this.sections[0]);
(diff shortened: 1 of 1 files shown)


dolanmiu:
Thanks!

PR #304, opened 2019-04-08
Fix table cell margin type XML attribute

### LINKED ISSUE #303: Table Cell Margin type XML attribute incorrectly named
The type of a cell margin XML element is being incorrectly set as `w:sz` when it should be `w:type`.

brucehappy:
Change table cell margin type attribute form w:sz to w:type as per http://officeopenxml.com/WPtableCellMargins.php

Fixes #303


dolanmiu:
Nice catch, thanks!

PR #307, opened 2019-04-09
Optimize XML output

### LINKED ISSUE #306: Optimize XML output
Currently the XML output always includes open and close tags for every element, even if that element has no children.  Additionally, there are some elements, such as `w:rPr` and `w:pPr` which do not need to be included in the output at all if they have no children.  The underlying `xml` library does support producing elements that are a single tag, but only if the value is an object containing a single `_attr` key, or if the value is null.  Therefore, the xml serialization should be changed to support sending objects in this format to the `xml` library for output.

Related to #213

brucehappy:
Rework the way attributes are stored in ImportedXmlComponent to match elsewhere (required allowing for a null xmlKeys in the XmlAttributeComponent interface).
Rework the way paragraphs get added to the end of table cells if needed.
The goal in both reworks is to not mess around with the objects output from `prepForXml` if we can avoid it.
Made the output of RunProperties, ParagraphProperties, TableCellProperties, TableRowProperties, and TableProperties all optional based on whether they contain any attributes or children.  Changed code in PageBorders, TableCellMargin, and TableCellBorders that implemented this same thing by overriding `prepForXml` so that it uses the new XmlComponent subclass instead.
Removed commented out code that attempted to fix-up XML output and make proper empty elements.
Fixed all affected tests.
Turn off `no-null-keyword` in the linter as we need to use null to signal to the `xml` library to create an empty element with no attributes (`undefined` will not work in its place).

Fixes #306

### DIFF
--- a/src/export/formatter.spec.ts
+++ b/src/export/formatter.spec.ts
@@ -31,7 +31,7 @@ describe("Formatter", () => {
             const paragraph = new file.Paragraph();
             paragraph.addRun(new file.TextRun("test").bold());
             const newJson = formatter.format(paragraph);
-            assert.isDefined(newJson["w:p"][1]["w:r"][0]["w:rPr"][0]["w:b"][0]._attr["w:val"]);
+            assert.isDefined(newJson["w:p"][0]["w:r"][0]["w:rPr"][0]["w:b"]._attr["w:val"]);
         });
 
         it("should format attributes (rsidSect)", () => {
--- a/src/export/packer/next-compiler.ts
+++ b/src/export/packer/next-compiler.ts
@@ -191,13 +191,4 @@ export class Compiler {
             },
         };
     }
-
-    /* By default docx collapse empty tags. <a></a> -> <a/>. this function mimic it
-       so comparing (diff) original docx file and the library output is easier
-       Currently not used, so commenting out */
-    // private collapseEmptyTags(xmlData: string): string {
-    //     const regEx = /<(([^ <>]+)[^<>]*)><\/\2>/g;
-    //     const collapsed = xmlData.replace(regEx, "<$1/>");
-    //     return collapsed;
-    // }
 }
--- a/src/file/content-types/content-types.spec.ts
+++ b/src/file/content-types/content-types.spec.ts
@@ -20,84 +20,70 @@ describe("ContentTypes", () => {
             expect(tree["Types"]).to.be.an.instanceof(Array);
 
             expect(tree["Types"][0]).to.deep.equal({ _attr: { xmlns: "http://schemas.openxmlformats.org/package/2006/content-types" } });
-            expect(tree["Types"][1]).to.deep.equal({ Default: [{ _attr: { ContentType: "image/png", Extension: "png" } }] });
-            expect(tree["Types"][2]).to.deep.equal({ Default: [{ _attr: { ContentType: "image/jpeg", Extension: "jpeg" } }] });
-            expect(tree["Types"][3]).to.deep.equal({ Default: [{ _attr: { ContentType: "image/jpeg", Extension: "jpg" } }] });
-            expect(tree["Types"][4]).to.deep.equal({ Default: [{ _attr: { ContentType: "image/bmp", Extension: "bmp" } }] });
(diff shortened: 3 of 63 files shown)


dolanmiu:
I would like to keep no-null-keyword on

dolanmiu:
Why have the tests changed so much, from `[{ ... }]` to `{ ... }`

bduncan-vm:
I commented on the most recent changes here:
https://github.com/dolanmiu/docx/issues/306#issuecomment-481543885

The main change (the reason the tests have changed) is in XmlComponent:
```
        // If we only have a single IXmlableObject in our children array and it
        // represents our attributes, use the object itself as our children to
        // avoid an unneeded XML close element.  (Note: We have to use this
        // function to get typescript to allow our check.)
        // Additionally, if the array is empty, use an empty object as our
        // children in order to get an empty XML element generated.
        const onlyAttrs = (c) => typeof c === "object" && c._attr;
        return {
            [this.rootKey]: children.length ? (children.length === 1 && onlyAttrs(children[0]) ? children[0] : children) : {},
        };
```

The tests changed from `[{ ... }]` to `{ ... }` because all those elements have no children.  As I mentioned in the comment linked above, and as you can see in the code snippet from `XmlComponent` above, there are there types of XML elements being sent to the `xml` library for output:

1. If you use `[ ... ]`, even if the array is empty, it means you want a start and end element in the XML
2. If you use `{ _attr : { ... } }` it means you want an empty element with attributes
3. If you use `{}` it means you want an empty element with no attributes

Anytime you see a test change from `[{ _attr: { ...} }]` to `{ _attr: { ... } }` it is because the element was being written as the first case (an element with children) and I have changed it to the second case (an element that does not have children, only attributes). Here's an example:

Old test would have looked like:
```
{ "w:bottom": [ { "_attr": { "w:type": "dxa", "w:w": "158" } } ] }
```
Old output would have been:
```
<w:bottom w:type="dxa" w:w="158">
</w:bottom>
```
New test looks like:
```
{ "w:bottom": { "_attr": { "w:type": "dxa", "w:w": "158" } } }
```
New output looks like:
```
<w:bottom w:type="dxa" w:w="158"/>
```

As for changing `{}` to a special token, sure, we could define and export a variable in the `xml-component.ts` file and use it in the single line of `XmlComponent.prepForXml` where `{}` is used, and in all the tests where `{}` appears.  Perhaps a better name would be `EMPTY_OBJECT` since it is an empty object, or `EMPTY_ELEMENT` since it produces an empty element in the XML.

The `IgnoreIfEmptyXmlComponent` class, which we only use for elements that do not need to be in the output if they are empty, such as `w:rPr`, returns `undefined` from `prepForXml` by not returning anything at all if the resultant object is empty.  This is the same thing you were doing in your `prepForXml` functions that checked if `this.root.length > 0`.  It results in the element being filtered out in its parent's `prepForXml` and therefore not going into the XML output at all.

dolanmiu:
Ok, understood

I am happy with either EMPTY_OBJECT or EMPTY_ELEMENT

If you make those changes, I will merge

Thanks again

bduncan-vm:
I've made the requested changes.  Let me know if there is anything else you need on this.

dolanmiu:
Thanks

I've added a comment on `protected readonly xmlKeys?:`

brucehappy:
Sorry, but I don't see any comments on `default-attributes.ts`:
https://github.com/dolanmiu/docx/pull/307/commits/816cb54b142a5518b700b815802d189b4d4c573d#diff-dea8b197f693d15f2c07e955494d22c0

The change was for here:
https://github.com/dolanmiu/docx/pull/307/commits/816cb54b142a5518b700b815802d189b4d4c573d#diff-087ff3fec208c1ae1d8122824e68366c
so that I could create the `ImportedXmlComponentAttributes` class and use it for  `ImportedXmlComponent` so that it could pass down the `_attr?: any` parameter from its constructor to make a set of attributes that do not need their names mapped to other names.  This allowed me to remove the `prepForXml` override in `ImportedXmlComponentAttributes`.

dolanmiu:
Thanks so much

Merged :)

PR #314, opened 2019-04-25
fix broken documentation links

efx:
@dolanmiu do you know how I can build and test the documentation locally? I didn't see anything in contribution guidelines. Perhaps we could document this process to make it easier for new contributors?


dolanmiu:
Ah yes, good point, I will add that in

It uses docsify, there is a `serve` command, but I can't remember on the top of my head

dolanmiu:
Thanks!

dolanmiu:
I have added a command called `npm run serve.docs`

PR #324, opened 2019-05-15
Fix merging of 3+ cells vertically

### LINKED ISSUE #321: Column MergeCells does not work over 3+ rows
The column cell merge needs to set the property of ALL of the cells in between startIndex and endIndex with 'continue', not just the last. 

I can't find the exact OpenXML reference for this, but it doesn't work right now (in 5.0.0.rc-1) and it does if I manually touch the XML to have the missing tags.

Scarface2013:
Fix #321


dolanmiu:
thank you,

ill release rc 2 soon

PR #337, opened 2019-06-20
Multiple columns in a section

zahrafali:


zahrafali:
addSection inserts stuff in a new page, is there a way we can have multiple sections in a single page?

dolanmiu:
Thanks! I will review this tomorrow

dolanmiu:
Looks good

The API may change though, but I will tweak this and update the demo before making another release

dolanmiu:
> addSection inserts stuff in a new page, is there a way we can have multiple sections in a single page?

I am not sure actually, will need to investigate

zahrafali:
> Looks good
> 
> The API may change though, but I will tweak this and update the demo before making another release
> 
> Thinking more of a:
> 
> ```ts
> doc.addSection({
>     column: {
>         width: 708,
>         count: 2,
>     }
> });
> ```

Yeah, this looks good

zahrafali:
> > addSection inserts stuff in a new page, is there a way we can have multiple sections in a single page?
> 
> I am not sure actually, will need to investigate

Okay, I will also investigate and see if I can fix it. What I've observed is, when we use addSection, it adds a new section to the document in a new page.

dolanmiu:
Released in `5.0.0-rc4`

PR #343, opened 2019-06-25
Make prettified XML a Packer-wide configurable option

markcharyk:
The xml library's `xml()` function was being passed a flag for pretty printing the XML in `word/document.xml`. This was causing some parsers (not Word itself) to break when they didn't expect whitepsace. It was also causing some files to be prettified and others to be ugly.

This ensures consistent prettification (and default non-whitespace XML) by taking an optional parameter at `Packer()` instantiation time that defaults to false and tells the compiler whether to compile in pretty (true) or ugly (false) XML.


dolanmiu:
Thanks!

dolanmiu:
Released in `5.0.0-rc4` https://github.com/dolanmiu/docx/releases/tag/5.0.0-rc4

Issue #388, opened 2019-08-26
Extra line before paragraph


jamesmontalvo3:
Thank you for a great module. I'm really looking forward to using it.

Using docx version `5.0.0-rc5` I'm getting an extra line prior to (or within) each paragraph. For example:

![image](https://user-images.githubusercontent.com/716482/63719450-8baf5180-c812-11e9-96da-b65def9a4a4f.png)

The above file was generated using the code below, which is a JS version of [demo/1-basic.ts](https://github.com/dolanmiu/docx/blob/d8b60d82f33b64b2c3223ee670713fe4f7d7517f/demo/1-basic.ts).

```js
var fs = require('fs')
var docx = require('docx');

const doc = new docx.Document();

doc.addSection({
	children: [
		new docx.Paragraph({
			children: [
				new docx.TextRun("Hello World"),
				new docx.TextRun({
					text: "Foo Bar",
					bold: true
				}),
				new docx.TextRun({
					text: "Github is the best",
					bold: true
				})
			]
		})
	],
});

docx.Packer.toBuffer(doc).then((buffer) => {
	fs.writeFileSync("docx-simple.docx", buffer);
});
```

bokuweb:
I run into same problem.

jamesmontalvo3:
Fixed by #391.

jamesmontalvo3:
Reopening due to #391 causing bigger issues and needing to be reverted. See #418.

dieterc:
Is there a work around?
I have the same issue and this is a bit annoying.

dolanmiu:
I think this is no longer an issue, I just tested this now and it doesn't seem to happen anymore

Closing.

Can re-open if its still persistant.

PR #391, opened 2019-08-29
fix: try to remove unnecessary paragraph

bokuweb:
I tried to update this library v4 to v5.
However I ran into extra line issue. https://github.com/dolanmiu/docx/issues/388
I tried to remove unnecessary paragraph, but I'm not sure if this behavior is intended.
If so, please reject this PR.

### DIFF
--- a/src/file/document/body/body.ts
+++ b/src/file/document/body/body.ts
@@ -1,5 +1,5 @@
 import { IXmlableObject, XmlComponent } from "file/xml-components";
-import { Paragraph, ParagraphProperties, TableOfContents } from "../..";
+import { TableOfContents } from "../..";
 import { SectionProperties, SectionPropertiesOptions } from "./section-properties/section-properties";
 
 export class Body extends XmlComponent {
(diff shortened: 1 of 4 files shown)


src/file/document/body/body.ts
```diff
@@ -18,9 +18,6 @@ export class Body extends XmlComponent {
      * @param options new section options
      */
     public addSection(options: SectionPropertiesOptions): void {
-        const currentSection = this.sections.pop() as SectionProperties;
-        this.root.push(this.createSectionParagraph(currentSection));
```
bokuweb: I guess this line generates empty paragraph.

```
    <w:body>
        <w:p/> <----- this
        <w:p>
            <w:r>
                <w:t xml:space="preserve">Hello World</w:t>
            </w:r>
            <w:r>
                <w:rPr>
                    <w:b w:val="true"/>
                    <w:bCs w:val="true"/>
                </w:rPr>
                <w:t xml:space="preserve">Foo Bar</w:t>
            </w:r>
            <w:r>
                <w:rPr>
                    <w:b w:val="true"/>
                    <w:bCs w:val="true"/>
                </w:rPr>
                <w:tab/>
                <w:t xml:space="preserve">Github is the best</w:t>
            </w:r>
        </w:p>
```

dolanmiu:
Thanks so much

I didn't like the reduction in coverage, but not sure why that happened

PR #407, opened 2019-09-26
Feat/declarative tables

dolanmiu:


jamesmontalvo3:
Did you intend to remove the capability to do:

```js
const table = new docx.Table({
	rows: 2,
	columns: 2
});
```

If so, why? This is so much simpler than:

```js
const table = new docx.Table({
	rows: [
		new docx.TableRow({
            children: [
                new docx.TableCell({
                    children: [],
                }),
                new docx.TableCell({
                    children: [],
                }),
            ],
        }),
        new docx.TableRow({
            children: [
                new docx.TableCell({
                    children: [],
                }),
                new docx.TableCell({
                    children: [],
                }),
            ],
        }),
    ]
});
```

What is the benefit of the latter? As a user of this module I'm just going to end up writing:

```
for (all rows) {
    create TableRow
    for (all columns) {
        create TableCell in TableRow
    }
}
```

This logic was in docx before.

dolanmiu:
> Did you intend to remove the capability to do:

Yes

> If so, why?

Because row:2, column:2 approach is not declarative, and relies on mutation

you would need to do:

```ts
table.getCell(0, 0).setChildren([])
table.getCell(0, 1).setChildren([])
table.getCell(1, 0).setChildren([])
table.getCell(2, 2).setChildren([]) // Would throw an exception because 2,2 doesn't exist
// I accidentally missed (1, 1)
```

>  This is so much simpler

Simpler yes, 100% agree, but simpler for the wrong reasons. The declarative, "complex" approach is much more clear, and does not rely on mutation (which IMO is bad).

If this is complex, then is HTML tables also complex?

https://www.w3schools.com/html/tryit.asp?filename=tryhtml_table

jamesmontalvo3:
Thinking about it more and actually implementing the necessary changes to my own code, I do prefer this new method. I was just surprised at the scope of the change. Do you expect other elements (besides tables) to have similar overhauls before releasing a stable 5.0.0?

Your point about HTML is taken, and in fact this change will make it significantly easier for me to abstract my DOCX writers from my HTML writers.

dolanmiu:
It's almost there

Now all there is left is the styles and numbering

So close!

PR #409, opened 2019-09-28
Simplify initiating table rows/columns

jamesmontalvo3:


jamesmontalvo3:
Can anyone explain why this fails tests? I see no functional difference between before/after code, just that the new code is simpler. Creating a simplified example, I see no difference between `tableRows1` and `tableRows2` below:

```javascript
const columns = 3;
const rows = 5;

class TableRow { constructor(cells) { this.cells = cells; } }
class TableCell { constructor() { this.x = 2; } }

let tableRows1 = Array(rows).fill(new TableRow(
	Array(columns).fill(new TableCell())
));


let tableRows2 = Array(rows)
	.fill(0)
	.map(() => {
		const cells = Array(columns)
			.fill(0)
			.map(() => new TableCell());
		const row = new TableRow(cells);
		return row;
	});

console.log(tableRows1);
console.log(tableRows2);
```

dolanmiu:
Because of this JavaScript "feature":

https://stackoverflow.com/questions/38940576/javascript-why-array-prototype-fill-actually-fills-a-pointer-of-object-when
https://stackoverflow.com/questions/37949813/array-fillarray-creates-copies-by-references-not-by-value

That's why it was done like that in the first place, to circumvent this

dolanmiu:
should probably add a comment in there to let people know why

PR #413, opened 2019-10-01
Add SymbolRun to allow adding symbols inline

jamesmontalvo3:
This allows adding symbols by doing the following:

```ts
const paragraph = new Paragraph({
    children: [
        new TextRun("This is an empty checkbox: "),
        new SymbolRun("F071"),
    ],
});
```

As shown in Word:

image

If you want to specify the symbol font you can do:

```ts
const paragraph = new Paragraph({
    children: [
        new TextRun("This is an empty checkbox: "),
        new SymbolRun({ char: "F071", symbolfont: "Arial" }),
    ],
});
```

### DIFF
--- a/docs/usage/paragraph.md
+++ b/docs/usage/paragraph.md
@@ -2,7 +2,7 @@
 
 > Everything (text, images, graphs etc) in OpenXML is organised in paragraphs.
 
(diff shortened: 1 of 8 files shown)


dolanmiu:
Perfect! good work, you're a star (or symbol)

PR #415, opened 2019-10-01
Export RunFonts to be able to pass to Level.addRunProperty()

jamesmontalvo3:
Exporting `RunFonts` allows this to be done:

```javascript
const level = abstract.createLevel(1, 'decimal', 'q');
level.addRunProperty(new docx.RunFonts('Wingdings'));
```

And gives output like:

image


dolanmiu:
Thanks!

PR #416, opened 2019-10-02
Fix table of contents \t flag

kalda341:
Use comma separated doublets (as described here: http://officeopenxml.com/WPtableOfContents.php) as opposed to semicolon separated.


dolanmiu:
Thanks!

PR #436, opened 2019-10-18
Support footnotes on TextRun elements

### LINKED ISSUE #432: Support footnotes on TextRun
## Feature Request

This is a feature request for supporting footnote references on the TextRun element.

## Usecase: 

We ran into the problem, that Paragraphs with mutliple text parts cannot be referenced by footnotes. It would be super handy, to being able to use the `referenceFootnote` function on a TextRun element.

**PS**: Thank you for creating this awesome lib. Works really well 🙌

zaunermax:
This PR implements the possibility to reference footnotes on TextRun elements.

closes #432

### DIFF
--- a/demo/17-footnotes.ts
+++ b/demo/17-footnotes.ts
@@ -1,14 +1,20 @@
 // Footnotes
 // Import from 'docx' rather than '../build' if you install from npm
 import * as fs from "fs";
-import { Document, Packer, Paragraph } from "../build";
+import { Document, Packer, Paragraph, TextRun } from "../build";
 
 const doc = new Document();
 
 doc.addSection({
-    children: [new Paragraph("Hello World").referenceFootnote(1), new Paragraph("Hello World").referenceFootnote(2)],
+    children: [
+        new Paragraph({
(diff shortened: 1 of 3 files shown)


jamesmontalvo3:
@dolanmiu may prefer you to write this declaratively, so rather than adding a method to TextRun objects, instead pass in footnote IDs along with the `ITextRunOptions` object used to instantiate TextRuns.

dolanmiu:
Thanks!

dolanmiu:
> @dolanmiu may prefer you to write this declaratively, so rather than adding a method to TextRun objects, instead pass in footnote IDs along with the `ITextRunOptions` object used to instantiate TextRuns.

Yes

I will refactor this for sure before releasing the next version

zaunermax:
Oh yeah, I just read the issues and discussions on how to keep this lib more declarative, which i generally like! I just wanted to keep it consistent with the paragraph element so I didn't look any further than that. 

Future PRs will keep in mind that declarative > imperative style especially in this usecase.

If you want I can try to transform the API to a declarative style if you want.

dolanmiu:
> If you want I can try to transform the API to a declarative style if you want.

Sure, if you can change it that would be great!

PR #444, opened 2019-10-30
Allow image reuse

mloar:
If you attempt to reuse an image within a document, all but the first instance of the image will be broken, because subsequent instances do not have their r:embed attribute rewritten to use the sequential ID.

This uses a regex with the global flag set in order to rewrite the r:embed attribute for every instance of the image.


dolanmiu:
Thanks, good spot

Can you fix the build?

Issue #440, opened 2019-10-25
By adding a hyperlink with doc.createHyperlink document won't open


ramonmata:
Hello,

If I run the code below or even with **demo/35-hyperlinks.ts**, the result word document can not be opened by word, resulting in a message saying:

> Word found unreadable content in links. Do you want to recover the contents of this document? If you trust the source of this document, click Yes.

I have manage to fix it by updating the logic in [relationships.ts](https://github.com/ramonmata/docx/blob/104c4c4deb84e1a5deb47ac7fc4dc156632f4cc6/src/file/relationships/relationships.ts#L27) which at this time seems more like a hack.

**Sample program to reproduce error**
```
import * as fs from "fs";
import { Document, Paragraph, TextRun, Packer} from "./build";


const doc = new Document();

doc.addSection({
    properties: {},
    children: [
        new Paragraph({
            children: [
                new TextRun('Documentation URL for '),
                doc.createHyperlink('https://docx.js.org', 'DOCX.js')
            ]
        })
    ]
});

Packer.toBuffer(doc).then((buffer) => {
    fs.writeFileSync("links.docx", buffer);
});
```

dolanmiu:
Fixed with 448

Issue #390, opened 2019-08-27
Adding Hyperlink breaks Word document 


dannycornelisseMT:
I'm on 5.0.0rc5. When I try to add a hyperlink to the document, I get the following message (while opening the word document): 'Word found unreadable content in "example (5).docx". Do you want to recover the contents of this document? If you trust the source of this document, click Yes.'

See the stackblitz below, note the versions of docx and file-saver in package.json

https://stackblitz.com/edit/angular-cny1zr?file=src/app/app.component.ts

Am I adding the hyperlinks wrong? 

Thanks in advance!

P.s. please excuse my not using a proper bug report template.

Edit, I'm on MacOS by the way

dannycornelisseMT:
@dolanmiu Is this something that is known and/or being worked on? Otherwise, I'll move on to an alternative solution.

p.s. Great library, awesome work

omgovich:
Hi, I have the same problem:

After adding a link to the document, MS Word can't open it without alert (tested it on Windows and macOS):
![photo5195323579934288889](https://user-images.githubusercontent.com/206567/65312641-9d7adf00-db9b-11e9-8af8-829d80db2e47.jpg)
 
I used the code from this example:
https://github.com/dolanmiu/docx/blob/master/demo/35-hyperlinks.ts

Package version: `5.0.0rc5`

Resulting document example: 
https://drive.google.com/file/d/1HlGzCMdbTVUZw2cP24hXgB9BycGkRG0T/

Sraleik:
same problem on version: `5.0.0-rc7`

```javascript
new Paragraph({
  children: [
    document.createHyperlink('https://www.google.com', 'Awesome link')
  ]
})
```

Is this the right way to create a link ?

Sraleik:
I found the problem 

in document.xml.rels my hyperlink as a duplicate id (here rId5) :

![duplicate_id](https://user-images.githubusercontent.com/9767889/68408949-92e5ca80-0186-11ea-9268-a2f77f4aefea.jpg)

i dont know how if it's the hyperlink that as the wrong relationshipsCount or if it's header || footer

but removind those duplicate ids fixes my problem

Sraleik:
#440 fix it for me too

dannycornelisseMT:
@Sraleik How did #440 fix it for you?

Sraleik:
as said by @ramonmata
It's a hacky fix. 
I'm trying to find a better one

but for now modifying the library like that works :

[...]

PR #448, opened 2019-11-08
fix hyperlink id

Sraleik:
I made this fix for problem with external hyperlink.

related to : #440 #390 

I Could not fix it by passing the proper relationshipCount so i just give a string id for this type of link

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   "dependencies": {
     "@types/jszip": "^3.1.4",
     "jszip": "^3.1.5",
+    "shortid": "^2.2.15",
     "xml": "^1.0.1",
     "xml-js": "^1.6.8"
   },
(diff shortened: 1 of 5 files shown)


dolanmiu:
Let me check what shortid does, 

I am concerned if all word processors support this, i assumed it had to be an integer number?

`rId0` 'rId6` etc

Sraleik:
I added shortid to fix this quickly. 
I need hyperlink for my Company project very quickly :)

But it could be a custom function that return an random string of n caracter.

i dont think a number is required for relation id, i'v seen `link1` `link2` from word generated document so i'v assumed that any string will do (i might be wrong)

dolanmiu:
Who do you work for? Mind if I add them to the readme?

dolanmiu:
Looks good to me, works fine, merging

Sraleik:
Thanks for the merge :+1: 

I work for Beekast : https://www.beekast.com/
Our current system doesn't use docxjs but next production delivery it will :)

dolanmiu:
Fixes https://github.com/dolanmiu/docx/issues/467

Issue #433, opened 2019-10-17
Empty paragraph at the start of document


olegdunkan:
When we crate simple document with only one paragraph:

```
doc.addSection({
    children: [
        new Paragraph({
            children: [new TextRun("Hello World")],
        }),
    ],
}); 
```
we see in resulting document before text "Hello World" empty line with line break
I exctracted from .docx file the document.xml and found out that it contains 

```
<w:body>
    <w:p/>
   <w:p>
      <w:r>
          <w:t xml:space="preserve">Hello world</w:t>
       </w:r>
    </w:p>
...
```
you can see <w:p/> is empty paragraph.

I have checked it out in code and I see 
`const currentSection = this.sections.pop() as SectionProperties;`

First call of this metod will generate empty <p/> element in document.xml.
https://github.com/dolanmiu/docx/blob/75c3c2f98500a891691047ae683d2fb15153ee59/src/file/document/body/body.ts#L21

Is it intened behaviour or bug?

jamesmontalvo3:
Related to #388 though it only appears to be happening at the beginning of the document now, not on every paragraph. It’s a bug.

PR #419, opened 2019-10-04
Revert "fix: try to remove unnecessary paragraph", which caused #418

jamesmontalvo3:
Fix broken multi-section documents.

- This reverts commit cb7486824764ac25ae5617e1a7028dd957376f02 / PR #391 
- Closes #418 
- Re-opens #388 (I personally had said #388 was resolved several days ago, but must not have tested against a multi-section document. Apologies.)

FYI @bokuweb

PR #476, opened 2019-12-15
Fix for empty first paragraph

mustapelto:
If there is only one section in a document, it always starts with an unnecessary empty paragraph, due to the way adding sections works. This fix simply removes the first element in the body's root array (i.e. the extra paragraph) during XML creation IFF there is only one section in the body. If there are multiple sections, it does nothing.

Will fix https://github.com/dolanmiu/docx/issues/433

### DIFF
--- a/src/file/document/body/body.spec.ts
+++ b/src/file/document/body/body.spec.ts
@@ -23,9 +23,6 @@ describe("Body", () => {
             expect(tree).to.deep.equal({
                 "w:body": [
                     {
-                        "w:p": {},
-                    },
-                    {
                         "w:sectPr": [
                             { "w:pgSz": { _attr: { "w:w": 10000, "w:h": 10000, "w:orient": "portrait" } } },
                             {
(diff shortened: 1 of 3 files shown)


dolanmiu:
I think someone already tried this, and reverted:

Please take a look

https://github.com/dolanmiu/docx/pull/419

dolanmiu:
Original PR:

https://github.com/dolanmiu/docx/pull/391

mustapelto:
No, I saw that too and their change was different. It removed the pop/push during section generation, which removed the section properties paragraph entirely, which lead to all sorts of problems. My fix doesn't touch the section properties at all.

```
public addSection(options: SectionPropertiesOptions): void {
    const currentSection = this.sections.pop() as SectionProperties;   //PR #391 removed this line
    this.root.push(this.createSectionParagraph(currentSection));        // and this one

    this.sections.push(new SectionProperties(options));
}

public prepForXml(): IXmlableObject | undefined {
    if (this.sections.length === 1) {
        //My fix adds a splice of this.root's first element here:
        this.root.splice(0, 1);
        this.root.push(this.sections.pop() as SectionProperties);
    }

    return super.prepForXml();
}
```

dolanmiu:
Thanks! Good work, will merge

PR #527, opened 2020-04-17
Enabled dark and light mode with switch for docs using docsify-darklight-theme with redesigned search bar

boopathikumar018:
Enabled dark and light mode with switch for docs using docsify-darklight-theme with redesigned search bar. For more customization [see here](https://boopathikumar018.github.io/docsify-darklight-theme)

**Light Mode :**

Before 

image

After

image

**Dark Mode :**

image


dolanmiu:
Wow, very nice, will review

PR #540, opened 2020-05-10
adds textDirection to table cells

bschwarz:
This adds _textDirection_ option to TableCell. There are three values for the _textDirection_ option (from the enumeration):

> BOTTOMTOTOPLEFTTORIGHT = "btLr",
> LEFTTORIGHTTOPTOBOTTOM = "lrTb",
> TOPTOBOTTOMRIGHTTOLEFT = "tbRl"

So, an example usage would look like this:
```
new TableCell({
    children: [new Paragraph({text: "top to bottom"}), new Paragraph({})],
    textDirection: TextDirection.TOPTOBOTTOMRIGHTTOLEFT
}),
```

Note that there are more options available from official [microsoft pages](https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.textdirectionvalues?view=openxml-2.8.1). However, the ones not implemented end in _*2010_ and _*rotate_. I wasn't sure if version specific options should be supported or not. And the *rotate didn't seem to make any difference when I tested. Of course, if you feel they all should be supported, then I am ok with that.  Also, I just uppercased the options from MS official. It's ugly, but I was trying to remain consistent. 

I also added a test case, and updated the table demo (31).

Tested on MicroSoft Word 365
Node version: v12.14.0

### DIFF
--- a/demo/31-tables.ts
+++ b/demo/31-tables.ts
@@ -18,12 +18,12 @@ const table = new Table({
                     verticalAlign: VerticalAlign.CENTER,
                 }),
                 new TableCell({
-                    children: [new Paragraph({text: "bottom to top"}), new Paragraph({})],
-                    textDirection: TextDirection.BOTTOMTOTOPLEFTTORIGHT
+                    children: [new Paragraph({ text: "bottom to top" }), new Paragraph({})],
+                    textDirection: TextDirection.BOTTOM_TO_TOP_LEFT_TO_RIGHT,
                 }),
                 new TableCell({
-                    children: [new Paragraph({text: "top to bottom"}), new Paragraph({})],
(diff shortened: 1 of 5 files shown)


dolanmiu:
Thanks, I have reviewed this and it looks good

Added a commit on top to improve the readability of the enums

dolanmiu:
Will release a new version of `docx` tomorrow with these changes

bschwarz:
Thanks for getting that in quick. Just a question (I am not trying to rush you), do you have an idea when you will publish into npm (I didn't see the new version in npm)? Just curious, no rush. I could patch my local copy, but I'd rather stick to the package manager.

dolanmiu:
it was published a few days ago 5.1.1

PR #547, opened 2020-05-27
Add appendRun method to Paragraph class

askoufis:
There used to a method called `addRun` on the `Paragraph` class but it got removed in the 5.0.0 release. It was useful and IMO an unnecessary breaking change. I called it `appendRun` as that name implies that it will be the last child and the breaking change has already been made so I figured now is a good time to rename it.

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -1,7 +1,7 @@
 {
   "check-coverage": true,
   "lines": 93.53,
-  "functions": 89.63,
+  "functions": 89.62,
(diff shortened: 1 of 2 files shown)


dolanmiu:
Nope

It was removed on purpose, deliberately

This library is declarative, so there will be no mutation. E.g. you can't create a `new Paragraph()`, then change and add stuff to it later

Remember: Think declarative

PR #564, opened 2020-07-15
Update table-row.ts to use correct height property

SoapyD:
the height property in table-row was set to height.height instead of height.value. This was then passing an undefined value to the setHeight function.

Corrected the issue above so the height property should work now.

### DIFF
--- a/src/file/table/table-row/table-row.ts
+++ b/src/file/table/table-row/table-row.ts
@@ -34,7 +34,7 @@ export class TableRow extends XmlComponent {
         }
 
         if (options.height) {
-            this.properties.setHeight(options.height.height, options.height.rule);
+            this.properties.setHeight(options.height.value, options.height.rule);
         }
     }
 


dolanmiu:
I do not think this is true, I see the error:

```
[at-loader] ./src/file/table/table-row/table-row.ts:37:54 
      TS2339: Property 'value' does not exist on type '{ readonly height: number; readonly rule: HeightRule; }'.
```

It says `value` does not exist

SoapyD:
Ah nuts! Apologies, i'm new to node and pull requests. In my local instance of package "docx": "^5.2.2", i amended docx/build/index.js row 17750 from this:

`        if (options.height) {
            this.properties.setHeight(options.height.height, options.height.rule);
        }
`
to this

`        if (options.height) {
            this.properties.setHeight(options.height.value, options.height.rule);
        }
`

the height property didn't work in the table until I did the above change; thought apply it to the src file would have to same affect. 

Think I may have to see how node packages build before any re-submission.

PR #565, opened 2020-07-16
table-row height correction

SoapyD:
resubmitting a pull request I sent on 15/07/2020. Cloned the project and followed the correct build procedure this time to hopefully ensure nothing bugs out when **dolanmiu** tries the build again.

Thought it best to convert instances of **height** to **value** as the xmlKey is mapping to **w:val** (which makes more sense to me) and the documentation refers to value as a property.

### DIFF
--- a/src/file/table/table-row/table-row-height.ts
+++ b/src/file/table/table-row/table-row-height.ts
@@ -10,12 +10,12 @@ export enum HeightRule {
 }
 
 interface ITableRowHeight {
-    readonly height: number;
+    readonly value: number;
     readonly rule: HeightRule;
 }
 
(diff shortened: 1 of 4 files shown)


dolanmiu:
> Thought it best to convert instances of height to value as the xmlKey is mapping to w:val (which makes more sense to me)

I like this, I feel the same way, good enough reason to me

But this is a breaking change, so it will be merged in when it's time to make v6

SoapyD:
Awesome :) first successful pull request! Will happily wait for v6.

dolanmiu:
Merged!

This will be in for V6

PR #596, opened 2020-07-31
feat: Add decimalFullWidth numbering type

SpringMT:
I want to use fullwidth number for numbering like `１` `２` . (I'm Japanese native)


dolanmiu:
Thanks @SpringMT 

This is released in 5.3.0

PR #224, opened 2018-12-20
Added trackChanges key to settings

bre7:
Closes #223 

Not 100% sure about Settings. Is this the correct way to add the key ?

Issue #223, opened 2018-12-20
Enable "Track Revisions"


bre7:
I'll submit a PR soon.

Add `w:trackRevisions` flag under settings (https://github.com/dolanmiu/docx/blob/master/src/file/settings/settings.ts ???)

`<w:trackRevisions w:val="true"/>`
or
`<w:trackRevisions w:val="false"/>`

dolanmiu:
sweet, ok, sounds good to me

dolanmiu:
Track revisions is now supported

PR #642, opened 2020-09-23
add settings option to add trackRevisions

thojansen:
This is a PR for issue #223 and basically a replacement for the open PR #224. Similar to the existing settings methods, it just adds a new empty element `w:trackRevisions`. I am not sure if the yes/ no value from the original PR adds any value, but apparently MS Word adds an empty element when the setting is toggled and otherwise it is omitted. Also a public methods is somehow cleaner than passing that value through various properties/ constructors (with the side effect that the existing unit tests do not need to be changed).

So, in a nutshell this PR adds the option to call,
```
doc.Settings.addTrackRevisions()
```
which has the following effect on the document.

Screenshot 2020-09-23 at 11 34 55

### DIFF
--- a/src/file/settings/settings.spec.ts
+++ b/src/file/settings/settings.spec.ts
@@ -79,4 +79,47 @@ describe("Settings", () => {
             expect(keys[0]).to.be.equal("w:compat");
         });
     });
+    describe("#addTrackRevisions", () => {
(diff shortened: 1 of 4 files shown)


thojansen:
closing in favour of #644

PR #644, opened 2020-09-24
basic support for track revisions

thojansen:
Creating a new PR instead of #642 and moving to another branch. The implementation is far from complete in terms of the specification for track revisions, but basic insertions and deletions are supported via new primitives. Any feedback is appreciated.

### Add setting to enable track revisions

This is a PR for issue #223 and basically a replacement for the open PR #224. Similar to the existing settings methods, it adds a new empty element `w:trackRevisions`. I am not sure if the yes/ no value from the original PR adds any value, but apparently MS Word adds an empty element when the setting is toggled and otherwise it is omitted. Also a public methods is somehow cleaner than passing that value through various properties/ constructors (with the side effect that the existing unit tests do not need to be changed).

So, in a nutshell this PR adds the option to call,
```
doc.Settings.addTrackRevisions()
```
which has the following effect on the document.

Screenshot 2020-09-23 at 11 34 55

## Add InsertedTextRun and DeletedTextRun

Add the option to create instances of `DeletedTextRun` and `InsertedTextRun` inside a paragraph, whereas the first one wraps an instance of a TextRun and the second one renders text as a `w:delText` node.

```
new InsertedTextRun({ child: new TextRun({ text: `xxx` }), id: 0, author: "author name", date: "2020-09-23T13:57:00Z" })
new DeletedTextRun({ text: 'xxx', id: 1, author: "author name", date: "2020-09-23T13:57:00Z"})
```


dolanmiu:
Thanks for this

looks good

Can you create a demo under the `/demo` folder to demonstrate the feature?

Would be a bonus to add something to the `/docs` as well for documentation

thojansen:
Yes, I can do that. I will start integrating this within the next week into my app and can then wrap this up.

dolanmiu:
Perfect, once that's done I will merge

thojansen:
@dolanmiu I have changed the signature to match a standard `TextRun`, added a demo and also provided a small documentation for change tracking as well.

dolanmiu:
Looks good, will merge!

PR #757, opened 2021-01-22
Added more data types for JSZip to consider using

arran4:
JSZip supports more options.

I had an issue getting the pre-commit hook working with this too.

There will likely be updates for things other than just the "ImportDotx" function.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,7 @@ indent_style = space
 indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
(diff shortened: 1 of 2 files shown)


dolanmiu (approved):

dolanmiu:
Could you fix the prettier issues?

arran4:
@dolanmiu Done.. It modified a lot of files.. (Lets see the test results.) I also modified .editorconfig as it seems it was opinionated about the line-endings so I set editor config to match. Let me know if that was too much.

PR #763, opened 2021-02-03
Allow styles and headings with numbering

### LINKED ISSUE #612: There is a confict between style and numbering
When I  set  a  numbering and style for the same paragraph, only the numbering works, the style I  write in style.js  is disappeared, this problem could be solved?
 
eg: 
      	return eachParagraph = new Paragraph({
				          pageBreakBefore:true,
				          style:"heading_1",          **dose not work**
				           numbering: {
				                reference: "my-crazy-reference",
				                level: 0,                      
				            },
			 	         });

netbymatt:
This allows for paragraphs to have numbering and not loose the ability to apply styles or headings, but still falls back to the default ListParagraph style for backwards compatibility. It addresses #612 and implements the change suggested there by @hjj1227.

Closes #612

### DIFF
--- a/src/file/paragraph/paragraph.spec.ts
+++ b/src/file/paragraph/paragraph.spec.ts
@@ -596,6 +596,36 @@ describe("Paragraph", () => {
             });
         });
 
+        it("should add a style to the list paragraph when provided", () => {
+            const paragraph = new Paragraph({
+                numbering: {
+                    reference: "test id",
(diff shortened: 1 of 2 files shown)


dolanmiu:
I see, yes this makes sense, could you fix the Prettier issues, then ill merge in

dolanmiu:
Thanks for this

PR #928, opened 2021-05-04
Added support for w:fldSimple, including convenience for MailMerge

pauldegroot:
### Simple fields
This adds support for SimpleFields, which are found in Office Open XML documents as w:fldSimple. These are fields like `FILENAME` or `TITLE`.

You can use these within paragraphs just like in TextRuns, like this:
```typescript
const doc = new Document({
    sections: [{
        properties: {},
        children: [
            new Paragraph({
                children: [
                    new TextRun("This file is called "),
                    new SimpleField("FILENAME", "Example.docx"),
                    new TextRun("!"),
                ]
            })
        ],
    }],
});
```

### Convenience class: SimpleMailMergeField
A convenience class was also added to include the fields for a mail merge. This automatically creates a cached value that make the field look as if we created this field using Microsoft Word.

Basically, the code:

```typescript
new SimpleMailMergeField("Name")
```

is a shorthand for

```typescript
new SimpleField("MERGEFIELD Name", "«Name»")
```

### DIFF
--- a/demo/66-fields.ts
+++ b/demo/66-fields.ts
@@ -1,0 +1,43 @@
+// Use fields to include dynamic text
+// Import from 'docx' rather than '../build' if you install from npm
+import * as fs from "fs";
+import { Bookmark, Document, Packer, Paragraph, SimpleField, TextRun } from "../build";
+
+const doc = new Document({
+    creator: 'Me',
+    sections: [
+        {
+            properties: {},
+            children: [
+                new Paragraph({
+                    children: [
+                        new TextRun("This document is called "),
+                        new SimpleField("FILENAME", "My Document.docx"),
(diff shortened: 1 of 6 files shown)


dolanmiu:
Nice work!

Could you create a file under `/demo` and write some documentation under `docs` about it?

pauldegroot:
Sure! Hope this matches your expectations.

dipakparmar:
@pauldegroot @dolanmiu do you know patching document with mail-merge fields is supported? I tried replacing `{{` and `}}` with `«` and `»` but it doesn't seem to work. 

https://github.com/dolanmiu/docx/blob/022b25cfcd33cd2ab9456f4b16cbc01ec3a1ed63/src/patcher/from-docx.ts#L130

PR #943, opened 2021-05-17
Fixes #877: table columns defaulting to small/unusable sizes in some word processors

### LINKED ISSUE #877: Table styles don't work in online word
Online word has problems with custom table styles.
for example the https://github.com/dolanmiu/docx/blob/master/demo/25-table-xml-styles.ts demo produces this in word:

image

but this in online word:

image

---
tested with `docx-6.0.3`

devoidfury:
Fixes #877 - don't add `w:tblGrid` to table when `columnWidths` argument isn't provided. Fixed the Table tests to account for this, and added one additional test with `columnWidths` set to make sure it's added in that case.

### DIFF
--- a/src/file/table/table.spec.ts
+++ b/src/file/table/table.spec.ts
@@ -178,9 +178,6 @@ describe("Table", () => {
             expect(tree).to.deep.equal({
                 "w:tbl": [
                     { "w:tblPr": [DEFAULT_TABLE_PROPERTIES, BORDERS, WIDTHS] },
-                    {
-                        "w:tblGrid": [{ "w:gridCol": { _attr: { "w:w": 100 } } }, { "w:gridCol": { _attr: { "w:w": 100 } } }],
-                    },
                     { "w:tr": [cell, cell] },
                     { "w:tr": [cell, cell] },
                     { "w:tr": [cell, cell] },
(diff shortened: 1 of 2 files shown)


dolanmiu:
What happens if no column width is provided? Does the document still look good?

devoidfury:
Yes, it automatically grows based on content and space, tested in Word Online.

Before:
Before screenshot

After:
After screenshot

devoidfury:
Hm, from what I can tell, in older versions of the standard it was a required element, but in the newer versions it's optional. So this may cause an issue with word 2007 or older -- worth a test there first.

devoidfury:
Looking at the ooxml wml.xsd, it looks like it's required; so probably shouldn't remove it.

```xml
  <xsd:complexType name="CT_Tbl">
    <xsd:sequence>
      <xsd:group ref="EG_RangeMarkupElements" minOccurs="0" maxOccurs="unbounded"/>
      <xsd:element name="tblPr" type="CT_TblPr"/>
      <xsd:element name="tblGrid" type="CT_TblGrid"/>
      <xsd:group ref="EG_ContentRowContent" minOccurs="0" maxOccurs="unbounded"/>
    </xsd:sequence>
  </xsd:complexType>
  ```

**Comment by devoidfury** [comment]
Also it says in the accompanying spec,

```
If the table grid is omitted, then a new grid shall be constructed from the actual contents of the table assuming
that all grid columns have a width of 0.
```

So the spec says one thing, but the schema says another. Fun. The schema is what's used by tooling though, so should probably stick to that.

devoidfury:
As of now, I don't think this is a good approach because in the xsd schema, it appears `w:tblGrid` is a required element in `w:tbl`.

dolanmiu:
@devoidfury Ok, if so, can this be closed?

PR #994, opened 2021-06-16
wrong border property

yaniv14:
based on IBorderPropertyOptions available properties are: color, space, value, size


dolanmiu:
Ok, i'll merge this in, I think the codecov is stuck, not sure why

dolanmiu:
correction

This is corrct, `style` will be used in `v7`

PR #1031, opened 2021-07-08
Export new `ColumnBreak` class via more generic `Break` class

askoufis:
The only new feature in this PR is the new `ColumnBreak` class. See [this spec](http://officeopenxml.com/WPtextSpecialContent-break.php) for more details.

I figured it made sense to extend the existing `Break` class by allowing a break type to be passed in as an enum. I didn't add the `textWrapping` break type as that is the default break type and is already available as part of the `Run` class.

This is my first time submitting a non-trivial feature to this library, so let me know if there's anything to improve for next time to make the PR process quicker.

This makes it possible to generate column layouts like this (this is the output of the demo file I added):
Screen Shot 2021-07-08 at 17 45 22


dolanmiu:
Perfect, looks good

PR #1057, opened 2021-07-26
Numbering start from number doesn't work fix

anti-the-social:
Use level.start for overrideLevel.start if possible.
Implemented via additional reference config map.


dolanmiu:
Could you have a look at the broken tests? @anti-the-social

dolanmiu:
@anti-the-social does demo 3 still work as expected?

https://github.com/dolanmiu/docx/blob/master/demo/3-numbering-and-bullet-points.ts

anti-the-social:
@dolanmiu just checked: result for demo number 3 at the master branch and "Numbering-start-from-number-doesn't-work" branch are identical
image
image

dolanmiu:
Perfect, merging

Issue #1119, opened 2021-08-27
ExternalHyperlink with multiple children


rowanc1:
I am trying to have multiple `TextRun` children be in the same ExternalHyperlink. The way that the API is designed, there is only a single `child`. This means that if there is formatting inside of a link (e.g. bold or other markup) you have to put the link on every single `TextRun`, this isn't ideal when creating links which people will then edit and/or remove in Word.

Single Hyperlink around two TextRuns:
![image](https://user-images.githubusercontent.com/913249/131196501-630fbbf5-f644-488c-a3c0-a6f905a64ef3.png)

It looks like the spec can support multiple children in a Hyperlink, it would be great if we could expose that in `docx`! I will put a draft PR together to show where the approximate changes are.

anti-the-social:
Just some example on what happens when you choose differently styled text and convert it to a link in Word
![image](https://user-images.githubusercontent.com/69673750/131467755-815182ed-fe0d-4885-b5ff-1cc1d1900932.png)
![image](https://user-images.githubusercontent.com/69673750/131467441-aad35570-7517-472f-83b1-d611f57aa6cd.png)

rowanc1:
Thanks @anti-the-social, I should have been more clear that the screenshot that I posted was from word, and created through the linked PR, applying the `Hyperlink` style as well as a few others in a different text run! Thanks for posting the code screenshot as well!

rowanc1:
We can likely close this issue now that #1120 has been merged in!

dolanmiu:
Closing as fixed by #1120

PR #1120, opened 2021-08-27
ExternalHyperlink with multiple children

rowanc1:
I am trying to have multiple `TextRun` children be in the same ExternalHyperlink. The way that the API is designed, there is only a single `child`, however, I have modified and tested multiple children on a word doc locally, and the hyperlink could potentially be extended to take multiple `TextRun`s?

This PR is not backwards compatible, and more work/advice would be needed to get it into shape. Looking for input to see if it is worth adding something along these lines?

See also #1119.

### DIFF
--- a/demo/21-bookmarks.ts
+++ b/demo/21-bookmarks.ts
@@ -18,10 +18,12 @@ const doc = new Document({
                         new Paragraph({
                             children: [
                                 new InternalHyperlink({
-                                    child: new TextRun({
-                                        text: "Click here!",
-                                        style: "Hyperlink",
-                                    }),
+                                    children: [
+                                        new TextRun({
+                                            text: "Click here!",
+                                            style: "Hyperlink",
+                                        }),
+                                    ],
                                     anchor: "myAnchorId",
                                 }),
                             ],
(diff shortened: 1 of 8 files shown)


dolanmiu:
Sure thing, will have a look at this PR

Thanks!

rowanc1:
Fantastic, let me know if I can do anything else here! Great library. :)

dolanmiu:
Can you update `demo/35-hyperlinks.ts`?

It is failing

rowanc1:
I have updated the demo and added an example of multi-children link.

Demo 35:
image

It looks like there is actually documentation for this feature, but it isn't linked into the table of contents? Also, it might be out of date? I could probably take a stab at updating that?

dolanmiu:
Sure yes, if you can update that documentation, that would be doubly appreciated!

rowanc1:
Nice will make me learn about internal hyperlinks now ... :). Will get this done in the next 30min or so.

rowanc1:
I think that there is a similar capability for internal hyperlinks, if we are making a backwards incompatible change for v7, is it also worth updating that now? I haven't confirmed that this actually works, but I suspect it will. That is changing:

```ts
new InternalHyperlink({
    child: new TextRun({
        text: "Anchor Text",
        style: "Hyperlink",
    }),
    anchor: "myAnchorId",
})
```

To:
```ts
new InternalHyperlink({
    children: [
      new TextRun({
          text: "Anchor Text",
          style: "Hyperlink",
      }),
      new TextRun({
          text: "with style",
          bold: true,
          style: "Hyperlink",
      }),
    ],
    anchor: "myAnchorId",
})
```

Seems like we should from writing the docs ...!? Let me know.

dolanmiu:
Yes updating that would also be good (but test afterwards)

rowanc1:
Sounds good. Tests seem to be working locally, will update this and demo 21 as well. Thanks for the fast response!

dolanmiu:
Perfect!

rowanc1:
Passing locally, demos 21 & 35 updated. Docs match the style of docs online and added a link in the contents (after bullet points).

image

rowanc1:
Let me know if there is anything else to do on this. Off for my eve, but can pick stuff up tomorrow.

dolanmiu:
Merged, looks good!

rowanc1:
Thanks! Looking forward to this being released! 🚀

PR #1131, opened 2021-09-01
Allow any `ParagraphChild` in `Bookmark`s

rowanc1:
This is a very minor change in types, I got there by adding other things into bookmarks. I am not sure if the typing is quite right, but this fixes type-bugs like:

image

Let me know if (and where) to do digging on what is actually allowed to be in a bookmark. I have put it to `ParagraphChild` as the children of a bookmark in this PR.


dolanmiu:
Thanks, merging, good catch

PR #1168, opened 2021-09-18
Extend `Columns` API to allow specifying individual column widths/spacing

### LINKED ISSUE #1030: Add API for specifying individual column widths/spacing
# Issue

Currently the API has a way to specify whether or not columns have an [equal width](https://github.com/dolanmiu/docx/blob/master/src/file/document/body/section-properties/properties/columns.ts#L37) or not. However, there is no way to specify the space after and width of the individual columns if you want them to be different widths.

Columns spec: http://officeopenxml.com/WPsectionCols.php

# Proposed API
To avoid a breaking change, an extension to the existing `column` section property seems like the best approach. I was thinking an optional field `columns` that takes a list of objects containing a `space` and `width`.

E.g.
```ts
const document = new Document({
  sections: [
    {
      properties: {
        column: {
          equalWidth: false,
          columns: [
            {
              width: 2835,
              space: 516,
            },
            {
              width: 5669,
            },
          ],
        },
      },
    },
  ],
});

```

I don't mind trying to implement this once a nice API is decided on.

askoufis:
Took a stab at implementing my suggestion from #1030. Let me know if there's anything I can improve.

Columns spec: http://officeopenxml.com/WPsectionCols.php

Closes #1030


dolanmiu:
demo 69, nice

dolanmiu:
I am about to release V7 soon, so we can add some breaking changes

what do you think about the following API?

```
columns: [
    new Column({ width: 2880, space: 720 }), 
    new Column({ width: 5760 })
],
```

The count is pretty obvious, as theres two elements in the array

askoufis:
I'll take another look and see if I can make it a bit easier to use with some breaking changes. Or would you rather it match the XML layout closely?

askoufis:
I'm not sure what happens if you have different `w:space` values on both `w:cols` and `w:col`, but it's possible to do with this API. There isn't any exact details about this situation, so I assume it's valid and that the `w:col` value overrides the parent `w:cols` value if present.

I feel like the API decent enough, and I'd rather keep the declarative style, so I'm happy with it.

PR #1163, opened 2021-09-16
Attaching numbering to custom style

anti-the-social:
Issue: https://github.com/dolanmiu/docx/issues/1147
Here are my first deep touches on docx :) Interesting structure, I should admit, took some time to learn it.

To sum PR up:
1) Moved numbering props level higher at `src/file/paragraph/properties.ts`
from `IParagraphPropertiesOptions` to `IParagraphStylePropertiesOptions`
2) Added numId replacer at `src/export/packer/next-compiler.ts` for styles.xml
3) Added demo on how it meant to be used at `demo/11-declaritive-styles-2.ts`

Didn't came up with some good tests since numbering is covered, while NumberingReplacer isn't in general.

PR #1172, opened 2021-09-20
Separate paragraph style options interface for level

anti-the-social:
Some polishing for this PR: https://github.com/dolanmiu/docx/pull/1163

The only place I found where IParagraphStylePropertiesOptions cannot have numbering property is level:
image

This PR would keep automatically created API documentation cleaner (https://docx.js.org/api/modules.html)
BUT This PR doesn't limit user to put numbering which would lead to unopenable file like so:
image


dolanmiu:
Thanks!

PR #1217, opened 2021-10-08
Change type of revision to a number

rowanc1:
I just worked through a bug that I didn't realize that `revision` in document properties had to be a number. I have changed the TypeScript to only allow numbers, which I think is clearer. This made the word doc angry when it was opened if there is arbitrary string in there.

image

The code change is backwards compatible as we cast this to a string before writing to XML. However, this will require people to unquote their revision numbers if they are using typescript.


rowanc1:
There could be another check in here to ensure that `options.revision >= 0`, negative numbers in there also complain.

Issue #1093, opened 2021-08-12
Word docs created don't work well in Dark Mode by default


jacwright:
When a Word doc is created in Microsoft Word, it has no background color defined. This allows it to work well in Dark Mode on devices that support it. docx adds a background color by default, and trying to override it with `null` or an empty string doesn't remove it.

We need a way to remove the background color and it really shouldn't have one by default. Here is a screenshot of 2 documents in Dark Mode. The left document was created by docx, the right document was created by Word. They both look normal in light mode, but they don't work well in dark mode. See how Word inverts the text color, but not the background color?

<img width="859" alt="Screen Shot 2021-08-12 at 11 41 21 AM" src="https://user-images.githubusercontent.com/120294/129244350-0da5af80-913e-47b8-ba05-c88f7d6060d8.png">

dolanmiu:
I thought this would be a simple feature, but it seems like it requires a big "dark mode" integration?

http://officeopenxml.com/prSlide-color.php

kptdobe:
Big +1.
And as a temporary workaround, if you need to manual "fix" the generated docx: go to `Design > Page Colour` and set `No Colour`.

dolanmiu:
Not something i'm working on, happy to review a pull request into this though

erlend-axelsson:
@dolanmiu I might be missing some big gotchas, but from my testing it seems like the pull request solves the dark mode issue.

PR #1285, opened 2021-11-12
Dont set background color if none is provided in options

erlend-axelsson:
After comparing the output of this library and what is
produced by Microsoft Word, it seems that document
background color is not set by default by Word.

Removing the default #FFFFFF background color enables the
generated documents to be readable in dark mode.

This seems to fix dolanmiu/docx#1093


dolanmiu:
Perfect. Merged

PR #1303, opened 2021-11-19
ability to specify textDirection for SectionProperties

katz:
This PR adds textDirection option to SectionProperties so that this library can generate vertical writing document.


katz:
I noticed that the similar feature, textDirection for tableCell, was introduced in #540.
This PR is intended to add an ability to create vertical writing at section (or page) lebel
vertical_writing_page
.

dolanmiu (approved):

dolanmiu:
Looks good!

PR #1323, opened 2021-11-30
Add an ability to specify charSpace and type for DocumentGrid

katz:
This PR adds  an ability to specify `charSpace` and `type` properties of `DocumentGrid`
DocumentGrid
.


dolanmiu:
Looks good!

PR #1345, opened 2021-12-18
#1342 add error if level exceeds

### LINKED ISSUE #1342: Add limit to numbering depth
MS Word and Google Docs both have a limit on how deep numbering and bullets can be. Here's a screenshot demonstrating that limit in Word

image

The same limit applies in Google Docs. This library allows creating numbering of arbitrary depth, but upon attempting to open the resulting document you get this error:

image

It took some time to figure out where the error was coming from, so it would be better if this library added a warning to the user that they've exceeded the maximum depth.

dolanmiu:
Closes #1342

### DIFF
--- a/docs/usage/bullet-points.md
+++ b/docs/usage/bullet-points.md
@@ -12,7 +12,7 @@ const doc = new Document({
             new Paragraph({
                 text: "Bullet points",
                 bullet: {
-                    level: 0 //How deep you want the bullet to be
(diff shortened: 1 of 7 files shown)


jamesmontalvo3:
It's going to be a couple weeks before I can look at this, but I just wanted to say how much I appreciate how quickly you addressed it. Thanks for such a great project!

PR #1384, opened 2022-01-20
Fix reading template images in the browser

### LINKED ISSUE #1272: Nodebuffer is not supported by this platform while using docxjs with javascript (not nodejs/react/vuejs).
I am trying to integrate docxjs to my website /browser. To Read the local file I use the below code. 

```  var xhr1= new XMLHttpRequest();
  xhr1.open('GET', 'template.dotx', true);
  if (xhr1.overrideMimeType) {
    xhr1.overrideMimeType('text/plain; charset=x-user-defined');
  }

  xhr1.onreadystatechange = function(e) {
    if (this.readyState == 4 && this.status == 200) {

      window.docData=this.response;
      }
  };
  xhr1.send();
```

**Called below function**
```
function generate() {

    const importDotx = new docx.ImportDotx();
     importDotx.extract(window.docData).then((templateDocument) => {
        const doc = new docx.Document({
            sections: [
                {
                    children: [
                        new docx.Paragraph({
                            children: [
                                new docx.TextRun("Hello World"),
                                new docx.TextRun({
                                    text: "Foo Bar",
                                    bold: true,
                                }),
                                new docx.TextRun({
                                    text: "\tGithub is the best",
                                    bold: true,
                                }),
                            ],
                        }),
                    ],
                },
            ],
        },                    
        {
            template: templateDocument,
        });
        
        
        
        docx.Packer.toBlob(doc).then((blob) => {
            console.log(blob);
            saveAs(blob, "example.docx");
            console.log("Document created successfully");
        });
        
    });
     

    }
```

I got below error Uncaught (in promise) Error: nodebuffer is not supported by this platform

simonbrunel:
[Media.addMedia()](https://github.com/dolanmiu/docx/blob/master/src/file/media/media.ts#L23) supports both node `Buffer` and `ArrayBuffer`, so check if we are running in the browser and use `ArrayBuffer`, instead of node `Buffer`, so we don't need to rely on a node buffer polyfill as [described in my comment](https://github.com/dolanmiu/docx/issues/1272#issuecomment-1017618628).

Fix #1272

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# https://prettier.io/docs/en/options.html#end-of-line
+* text=auto eol=lf
--- a/src/import-dotx/import-dotx.ts
+++ b/src/import-dotx/import-dotx.ts
@@ -157,7 +157,8 @@ export class ImportDotx {
         const hyperLinkReferences = this.findReferenceFiles(xmlRef).filter((r) => r.type === RelationshipType.HYPERLINK);
 
         for (const r of wrapperImagesReferences) {
-            const buffer = await zipContent.files[`word/${r.target}`].async("nodebuffer");
+            const bufferType = JSZip.support.arraybuffer ? "arraybuffer" : "nodebuffer";
+            const buffer = await zipContent.files[`word/${r.target}`].async(bufferType);
             const mediaData = media.addMedia(buffer, {
                 width: 100,
                 height: 100,


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# https://prettier.io/docs/en/options.html#end-of-line
+* text=auto eol=lf
```
simonbrunel, line 2: Enforces LF EOL when doing a checkout of this project. On Windows, `git commit` fails when `core.autocrlf: true` because of the pre-commit hook running prettier, which expects file with LF EOL. Let me know if you prefer to not add this file.

dolanmiu: nice one

PR #1440, opened 2022-03-03
Implement sample version of Comment feature.

chenyuncai:
Implement the comment feature.

1. add comment-run element.
2. add document attr config for comment feature.
3. add content-type and relationships item.
4. add a comment builder demo.
5. change the packer prettier param to string as sometime you can pass two blanks to xml output.

### DIFF
--- a/src/export/packer/next-compiler.ts
+++ b/src/export/packer/next-compiler.ts
@@ -5,6 +5,7 @@ import { File } from "file";
 import { Formatter } from "../formatter";
 import { ImageReplacer } from "./image-replacer";
 import { NumberingReplacer } from "./numbering-replacer";
+import { PrettityType } from "./packer";
 
 interface IXmlifyedFile {
     readonly data: string;
@@ -42,7 +43,7 @@ export class Compiler {
         this.numberingReplacer = new NumberingReplacer();
     }
 
-    public compile(file: File, prettifyXml?: boolean | string): JSZip {
+    public compile(file: File, prettifyXml?: boolean | PrettityType): JSZip {
         const zip = new JSZip();
         const xmlifiedFileMapping = this.xmlifyFile(file, prettifyXml);
         const map = new Map<string, IXmlifyedFile | IXmlifyedFile[]>(Object.entries(xmlifiedFileMapping));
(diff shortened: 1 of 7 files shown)


chenyuncai:
@dolanmiu Hi, there is PR about word comment feature.

src/file/paragraph/run/comment-run.ts
```diff
@@ -0,0 +1,155 @@
...
+
+interface ICommentOptions {
+    readonly id: string;
+    readonly initials?: string;
+    readonly author?: string;
+    readonly date?: string;
+}
+
+class CommentAttributes extends XmlAttributeComponent<ICommentOptions> {
+    protected readonly xmlKeys = { id: "w:id", initias: "w:initials", author: "w:author", date: "w:date" };
```
VictorBaron, line 13: typo here: `initias` --> `initials`

src/import-dotx/import-dotx.ts
```diff
@@ -37,11 +38,12 @@ interface IRelationshipFileInfo {
...
-    readonly headers: IDocumentHeader[];
-    readonly footers: IDocumentFooter[];
-    readonly styles: string;
-    readonly titlePageIsDefined: boolean;
-    readonly media: Media;
+    readonly headers?: IDocumentHeader[];
+    readonly footers?: IDocumentFooter[];
+    readonly styles?: string;
+    readonly titlePageIsDefined?: boolean;
+    readonly media?: Media;
```
dolanmiu, line 45: Why are these now optional?

src/export/packer/next-compiler.ts
```diff
@@ -64,7 +65,7 @@ export class Compiler {
         return zip;
     }
 
-    private xmlifyFile(file: File, prettify?: boolean): IXmlifyedFileMapping {
+    private xmlifyFile(file: File, prettify?: boolean | string): IXmlifyedFileMapping {
```
dolanmiu, line 68: It is now a boolean and a string, what could the string be used for?

dolanmiu: It seems like it's used like so:

```ts
Packer.toBuffer(doc, '  ').then((buffer) => {
  fs.writeFileSync("document-comments.docx", buffer);
});
```

I would prefer it if it is passed as an option in `File` instead

chenyuncai: Change the normal string to a enum type to protect the xml output content.

PR #1563, opened 2022-06-26
Recommend to install `Code Spell Checker` Extension

khill-fbmc:
Since I saw some [cSpell definitions in the settings](https://github.com/dolanmiu/docx/blob/master/.vscode/settings.json#L2=), why not also recommend to install the extension?


dolanmiu:
cSpell isn't installed actually, check the package.json

this was an experiment which didn't quite get finished. This is because there were too many false positives with stuff like `w:id`, which cSpell found errors with

PR #1574, opened 2022-07-06
docs: Fix property name accuracy in describing paragraph borders

wvbe:
Found this tiny inaccuracy in the docs. When giving a paragraph a border, the border style is determined by the `style` property, not the `value` property.


dolanmiu:
Thanks for this!

PR #1582, opened 2022-07-08
Update _sidebar.md

### LINKED ISSUE #1581: Accidental Menu Nesting?
It appears to me that some of these menu items should be in the top level and not sub-lists of bullet points

image

khill-fbmc:
Correct me if I am wrong, but like this? To fix #1581


dolanmiu:
Thank you! correct yes

PR #1592, opened 2022-07-13
Use strict on ts-config

dolanmiu:


dolanmiu:
Already done in another PR

PR #1633, opened 2022-08-19
Add toString to Packer

Volper212:
I found that having `toBase64String` but not `toString` is kinda weird, so I wrote it myself, which ended up the same as the other methods except for changing the `type` property.

Do I need to write a test for this to be approved? It's almost the same as the other methods and it works on my app. One thing I was concerned about was breaking the built-in JS `toString` function, however I don't think this class is converted to a string anywhere, is it?

### DIFF
--- a/src/export/packer/packer.spec.ts
+++ b/src/export/packer/packer.spec.ts
@@ -37,6 +37,14 @@ describe("Packer", () => {
         });
     });
 
+    describe("#toString()", () => {
+        it("should return a non-empty string", async () => {
+            const result = await Packer.toString(file);
+
+            assert.isAbove(result.length, 0);
+        });
+    });
+
     describe("#toBuffer()", () => {
         it("should create a standard docx file", async function () {
             this.timeout(99999999);
--- a/src/export/packer/packer.ts
+++ b/src/export/packer/packer.ts
@@ -13,6 +13,17 @@ export enum PrettifyType {
 }
 
 export class Packer {
+    public static async toString(file: File, prettify?: boolean | PrettifyType): Promise<string> {
+        const zip = this.compiler.compile(file, prettify);
+        const zipData = await zip.generateAsync({
+            type: "string",
+            mimeType: "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+            compression: "DEFLATE",
+        });
+
+        return zipData;
+    }
+
     public static async toBuffer(file: File, prettify?: boolean | PrettifyType): Promise<Buffer> {
         const zip = this.compiler.compile(file, prettify);
         const zipData = await zip.generateAsync({


dolanmiu:
Yup, you'd need to write some simple tests as the code now reduced the coverage

Volper212:
I added a simple test which increased the coverage of `packer.ts` to 100%.

I can see that there's another test for each of the other methods, however I didn't copy paste it because it just seems useless and I haven't seen it add any coverage. Does it really even test anything? I thought at first that it might be because it should be handled by `.catch` and not only by a regular try..catch block, however this is an async function so every error can be caught using either approach, AFAIK. Also, it says it "should handle" the exception which is misleading because the user has to handle it themselves.

dolanmiu:
Thanks looks good, merging

Regarding the tests, yes the tests could be better and more meaningful. Would be much appreciated if you could improve the tests

For me, I believe the hidden value of writing tests is to force the code to be written in a testable way, so even if the tests don't do much, the fact that it can even be tested is good.

Volper212:
Thanks for the merge.

I think you might have wanted to check the `.catch` vs try..catch with those tests in case someone wrote those methods using `.then` and not the async keyword (in that case the error wouldn't be caught using `.catch`), but it feels like this could be a linter/style thing (I just looked it up - the rule exists and it's [@typescript-eslint/promise-function-async](https://maximorlov.com/linting-rules-for-asynchronous-code-in-javascript/#14-typescript-eslint-promise-function-async)). This means that those tests could be deleted[^1]. I've seen that you use TSLint though (it's deprecated now), which may not have this rule implemented - maybe time for a migration to ESLint?

[^1]: I don't recommend deleting a passing test though, unless it's a noticable slowdown - instead do it when it fails after changing either the implementation details that the test stubs or the expected behavior of the function

Volper212:
Apparently, "string" is deprecated in JSZip in favor of "binarystring", but it works, so I guess it's OK.

dolanmiu:
> Thanks for the merge.
> 
> I think you might have wanted to check the `.catch` vs try..catch with those tests in case someone wrote those methods using `.then` and not the async keyword (in that case the error wouldn't be caught using `.catch`), but it feels like this could be a linter/style thing (I just looked it up - the rule exists and it's [@typescript-eslint/promise-function-async](https://maximorlov.com/linting-rules-for-asynchronous-code-in-javascript/#14-typescript-eslint-promise-function-async)). This means that those tests could be deleted[1](#user-content-fn-1-09ae742fcb0f86db4cf8576607c73c95). I've seen that you use TSLint though (it's deprecated now), which may not have this rule implemented - maybe time for a migration to ESLint?
> 
> ## Footnotes
> 1. I don't recommend deleting a passing test though, unless it's a noticable slowdown - instead do it when it fails after changing either the implementation details that the test stubs or the expected behavior of the function [↩](#user-content-fnref-1-09ae742fcb0f86db4cf8576607c73c95)

I want to use ESLint yes, but it seems like a huge effort, and TSLint works really well, (and I am a fan of Palantir, the company), so I don't see any urgency to update, but yes, would be nice if it was updated

dolanmiu:
> Apparently, "string" is deprecated in JSZip in favor of "binarystring", but it works, so I guess it's OK.

Right, so best to use binarystring before next version release?

Volper212:
Yes and you could change `Packer.toString` to `Packer.toBinaryString`

PR #1636, opened 2022-08-22
Create packer stream "toStream"

MLD5000:
This creates a new DOCX export function that returns a readable stream.  I created this to help reduce memory needs when generating a document with a log of image files.

I had to update the webpack config file because JSZip was not detecting the NodeJS stream object during testing and threw errors for jszip.generateNodeStream() function.


dolanmiu:
This might break browser compatability, but lets see

Issue #437, opened 2019-10-21
how to add horizontal line?


cool-firer:
is there a way to add horizontal line, cant find api related to it.

rajavijayavendan:
@cool-firer Any workaround for drawing shapes?

cool-firer:
@rajavijayavendan no solution.

nasriime:
try this works in version 5.0.0-rc5    

new Paragraph({
                    text: " ",
                    thematicBreak: true,
                }),

dolanmiu:
This is how you do it with `docx`. It add's a border to a paragraph

https://github.com/dolanmiu/docx/blob/master/demo/26-paragraph-borders.ts

This is how it is done officially with Microsoft Word too:

Word generated code:

```xml
<p w:rsidR="00F371A5" w:rsidRDefault="00F371A5">
	<pPr>
		<pBdr>
			<bottom w:val="single" w:sz="12" w:space="1" w:color="auto" />
		</pBdr>
	</pPr>
	<bookmarkStart w:id="0" w:name="_GoBack" />
	<bookmarkEnd w:id="0" />
</p>
```

PR #1677, opened 2022-09-26
Added the name horizontal line or horizontal rules

hunterwilhelm:
There was some confusion in https://github.com/dolanmiu/docx/issues/437 because the docs don't mention anything called horizontal lines or horizontal rules. 

Microsoft calls them by those names here https://support.microsoft.com/en-us/office/insert-a-line-9bf172f6-5908-4791-9bb9-2c952197b1a9 

This will make it easier for people to find this feature in the docs.

### DIFF
--- a/docs/usage/paragraph.md
+++ b/docs/usage/paragraph.md
@@ -111,7 +111,7 @@ const paragraph = new Paragraph({
 
 ## Border
 
-Add borders to a `Paragraph`. Good for making the `Paragraph` stand out
+Add borders to a `Paragraph`. Good for making the `Paragraph` stand out. Also called horizontal lines or horizontal rules.
 
 #### IBorderPropertyOptions
 


dolanmiu:
Borders and Horizontal Rule are semantically different though

I understand the point of your PR, so it's valuable, but maybe it can be worded a bit differently

Borders is like.... `border-top`, `border-left`, `border-right` in HTML
Whereas horizontal rule is the `<hr />` element

hunterwilhelm:
Good point, but that begs the question: How do you add a horizontal rule then? The solution in https://github.com/dolanmiu/docx/issues/437 was border top? 

How do we add the <hr /> element with dolanmiu/docx?

dolanmiu:
Thanks the the amendments

I did some googling too, but couldn't find anything about horizontal rules in OOXML, so maybe borders are the best we got. Will merge this PR once the build is finished

Thanks!

PR #1700, opened 2022-10-08
Add MathIntegral class

tvale1:
Added Math Integral class


dolanmiu:
Thank you!

I have fixed the prettier and linting issues, and wrote a test

When it passes, I will make sure to merge

Issue #1300, opened 2021-11-19
Multiple tabstops in a paragraph does not work properly with LibreOffice


mathoi:
Hi, 
I use version 7.1.2 of this library.

The documents I generate with this library looks really good in Microsoft Word, but in other editors the results are not always as good. One of the main issues seems to be multiple tabstops in a single paragraph (which I use a lot.)

I'm adding multiple tabstops to a paragraph using the property
 
 tabStops?: { leader?: LeaderType; position: number; type: TabStopType }[]

of IParagraphOptions, but in Libre Office (7.2.2.2) it only uses the last tabstop of the array, i.e. the first tab i add to the paragraph will use that tabstop and every tab after the first will be a default tabstop.

To make a quick test if Libre Office have difficulties with multiple tabstops, I have manually made a simple document with multiple tabstops in Microsoft Word and then opened it i Libre Office. It did not have any problems with the tabstops - the problems only occurs when I use this library to generate documents.

In the following is an example of document, that I've made with this library and how it looks in Microsoft Word 2010 vs Libre Office (7.2.2.2)

Here is how the document looks in Microsoft Word 2010 and how it is supposed to look:

![tabStops-word](https://user-images.githubusercontent.com/10826024/142600783-4d245d33-4285-429e-9441-66d92229e48c.png)

Here is how it looks in Libre office:

![tabStops-libreOffice](https://user-images.githubusercontent.com/10826024/142600802-13585a2a-c45f-470b-894e-a12c12aae59d.png)

anti-the-social:
Hi!
1) It would be much appreciated if you provide code to run to show the issue
2) 95% chance that issue isn't in the docx js library since LibreOffice does have its own logic which is not 100% following OOXML.
3) You can try to create file straight in LibreOffice in a way you like - then you'll test if LibreOffice is capable. Then save it as docx and try to open in Word. If outlook would be the same then problem is in docxjs.

My guess is that LibreOffice doesn't match Word logic and you would not be able to create same looking double tab stop in both of those programs.

mathoi:
Hi,
Thanks for the reply. Here are my comments for your 3 bullets

### Regarding 3 (and 2)
If I manually make a document i LibreOffice and then open it in Microsoft Word then it looks fine. The same is true for the other direction.
This suggests that it is possible to make tabstops that works correctly and similarly in both Mircrosoft Word and LibreOffice. This even seems to be the default case when making a document in either Microsoft Word or LibreOffice.
 

[...]

PR #1709, opened 2022-10-14
FIX: multiple tabStop support for LibreWriter

ronram5126:
Have faced issue with tabStop while using docx.js. Turns out docx generates multiple TabStops with single TabStop. This was not supported by LibreOffice.

So I forked and fixed the issue. I am currently using the fork in my project, will test and update if something else comes up.

Cheers,
Ronit R. BK

### DIFF
--- a/demo/75-tab-stops.ts
+++ b/demo/75-tab-stops.ts
@@ -4,19 +4,18 @@ import * as fs from "fs";
 import { Document, HeadingLevel, Packer, Paragraph, TabStopPosition, TabStopType, TextRun } from "../build";
 
 const columnWidth = TabStopPosition.MAX / 4;
-const reciptTabStops = [
-    // no need to define first left tab column
-    // the right aligned tab column position should point to the end of column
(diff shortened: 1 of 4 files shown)


ronram5126:
This fixes issue #1300

dolanmiu:
Thanks for the PR!

I have fixed the prettier and linting errors

I appreciate the effort you made to make it backwards compatiable in the TabStops, but theres no need for that, as it is not exposed to the user, so it will not break the API, we can modify no problem. And that is what I have done. It is much more simple now.

If you can write a demo using the tab stops, that would be great

dolanmiu (changes requested):
Would like you to add a demo to demonstrate this feature please under the `/demo` folder

ronram5126:
@codecov-commenter  I'm busy with work a bit, will do tomorrow. Cheers.

ronram5126:
I have added a demo/75-tab-stops.ts, I hope its fine. Let me know if you need something else.
Cheers.

PR #1725, opened 2022-10-21
Change `globalObject` to `globalThis`

kkocdko:
I want to run docs.js in some embed js engine which have not `this` defined in global namespace (and es module also haven't).

Compatibility: not a problem, `globalThis` was added in ES6, and this project's `tsconfig.json` also uses ES6 target.

https://github.com/dolanmiu/docx/blob/master/tsconfig.json#L3

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -12,7 +12,7 @@ const configuration = {
         filename: "index.js",
         libraryTarget: "umd",
         library: "docx",
-        globalObject: "this",
+        globalObject: "globalThis",
     },
 
     resolve: {


dolanmiu:
I am not sure about this, this will break `docx` on the web:

Docs from webpack:

https://webpack.js.org/configuration/output/#outputglobalobject

```
To make UMD build available on both browsers and Node.js, set output.globalObject option to 'this'
```

kkocdko:
> Webpack doc: To make UMD build available on both browsers and Node.js, set output.globalObject option to 'this'

I think this description is just for legacy. I have tried on Chrome, Firefox, Deno, Node, QuickJS and their ESModule enviroument. There's no reason that `globalThis` will break on web environment. And, 

> [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#description): `this` will be undefined in Modules

So the most important problem is compatibility. Chrome >= 71, Firefox >= 65, no IE supports. But I'm not sure what target `docx.js` supports before?

dolanmiu:
Ok, sounds good, I will double check tomorrow, if its good, I will merge

dolanmiu:
Looks good to me, merging

Issue #1740, opened 2022-10-26
lvlOverride logic adjustments


anti-the-social:
Moi :)

Currently concrete numbering related to overwriting starting value is applied to one level only.
<img width="383" alt="image" src="https://user-images.githubusercontent.com/69673750/198001148-2d5501b0-8164-4ffb-9650-da3c8a58a453.png">

Sometimes users want to start next concrete numbering with other level, than first one, but if we keep "overrideLevel" logic as it, numbering would be continued. Here is how it looks like: (`pref 1.2` meant to be `pref 1.1` since its restarted)

<img width="550" alt="image" src="https://user-images.githubusercontent.com/69673750/198002228-74aadd8e-273d-4cb9-a242-fae248703f23.png">

In order to fix this behaviour we have to be able to generate similar xml to this:
<img width="383" alt="image" src="https://user-images.githubusercontent.com/69673750/198001604-72920dc0-f6e4-4f66-94aa-c802630f459b.png">

And this is the reference to the place where its hardcoded:
https://github.com/dolanmiu/docx/blob/563e10fa92d206fdad897f243994386a5437e8a0/src/file/numbering/numbering.ts#L214

Potentially we can just propogate override for all levels with number 1, if instance has changed...But that would not be flexible e.g. you would be able to start with 10.1.1 after reset, but wouldn't be possible with 10.1.10

The best way, I guess, is to give extensive control over which level has to be actually reseted e.g. `[{lvl: 0, start: 1}, {lvl:1, start : 10}]`

dolanmiu:
Hello :)

Hmm yeah I see what you mean, may require significant change in API you reckon?

anti-the-social:
Yep, it does sound to me as a quite a break changes. 
`instance` at `IConcreteNumberingOptions` seem to be not enough to be "as flexible as possible"

Nevertheless, we have an option where instance would just set all levels to 1 and optional overrideLevel.start would modify first level only. That might stay compatible with old versions...

anti-the-social:
Which way we want to go?
I might do PR, but wouldn't promise it to be fast 😅
Lets see if that would be very critical to the project I am working on.

dolanmiu:
Sure thing

Maybe something like this could be cool:

```ts
const concreteNumbering = new ConcreteNumbering({
    numId: 0,
    abstractNumId: 1,
    reference: "1",
    instance: 0,
    overrideLevels: [{
        num: 1,
        start: 1,
    }, {
        num: 2,
[...]

PR #1823, opened 2022-12-07
Override levels impl

anti-the-social:
Bonjour :) 

Finally got some time to take care of this.
Tell me the secret of running tests! :)  Just after polling and `npm i ` you cannot really run `npm test`
I am getting `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for `

Nevertheless I believe it should work.
I also fixed and added additional tests to this **breaking change** improvements.

This pull request is related to this issue: https://github.com/dolanmiu/docx/issues/1740


anti-the-social:
Nanoid has critical vulnerability in the older versions, than 4.
I also have doubts that its the reason (just tried `npm i nanoid 3` `npm test`)

dolanmiu:
> Nanoid has critical vulnerability in the older versions, than 4. I also have doubts that its the reason (just tried `npm i nanoid 3` `npm test`)

No it must be, because there is an open PR which only upgrades nanoid from 3 -> 4, and it broke:

https://github.com/dolanmiu/docx/pull/1548

Take a look

dolanmiu:
There is one broken test in the project

anti-the-social:
Fixed failing test - my bad.

And also found after local testing that some tests at `src/file/paragraph/run/comment-run.spec.ts` were failing for me - reason seems to be the timezone. So I've fixed it via specifying full date in `new Date()` creation. 

Now its all good for me too
image

PR #1899, opened 2023-01-23
Feat/strict typescript

dolanmiu:


dolanmiu:
Work is superceeded by Vite upgrade

PR #2016, opened 2023-03-20
Images in templates

ewoutkleinsmann:
I ran into some issues when using the patchDocument function with a template that contains images. The function tries to parse the images as XML and (naturally) throws an error. I created a small fix for it. 

Would be great if you could merge this. Let me know if you want me to change something!


dolanmiu:
Oh derp, of course. Missed this case

Thank you!

PR #2087, opened 2023-04-18
fix: added unique numeric id creator to avoid numbering render errors

### LINKED ISSUE #2027: Lists numbers not rendered in 'My office'
I have simple doc, that looks find in word, but misses the list numbers in My Office.
word:
snapshot
My Office:
snoopshot

lentyaevpk:
There are issues with numbering render in MyOffice. It happens when `concreteNumbering` numIds look like 1,5,6, because we have the same `currentCount` for all numIds. And if `concreteNumbering` numIds are consecutive everything renders correctly.

Resolves #2027


dolanmiu:
Thanks, merging, will cut a release tonight

PR #2103, opened 2023-04-28
fix: do not use static numeric counters

VarLog:
Create new unique numeric Id creators on each file generation.

Using static shared instance method have an issue when counters growing up permanently on each document generation if you use docx library on nodejs server. And some docx editors do not show numeration properly when counters do not start from zero.

### DIFF
--- a/src/file/drawing/anchor/anchor.spec.ts
+++ b/src/file/drawing/anchor/anchor.spec.ts
@@ -1,8 +1,6 @@
 import { assert, expect } from "chai";
-import { SinonStub, stub } from "sinon";
 
 import { Formatter } from "@export/formatter";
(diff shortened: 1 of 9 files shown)


dolanmiu:
What would this do?

VarLog:
@dolanmiu Hello. Sorry for empty description, I just had to be sure that this fix will work.

So, we ran into a problem with numeration in some of the docx editors. We are using docx library on nodejs server and found out that only the first generated document works correctly in our target docx editor.

Following the investigation I found that the reason of it is non-zero numeric ids. And this pull request enforce create new unique numeric id counters on each document generation because counters will create in constructions now.

Thank you for your library and please accept this PR (if I did not break anything down, of course :) )

dolanmiu:
Released in `8.0.4`

Issue #2222, opened 2023-07-06
Packer.toBuffer() returns a UInt8Array


tripodsan:
with version 8.2, the `Packer.toBuffer()` now returns a UInt8Array:

![image](https://github.com/dolanmiu/docx/assets/917628/b5785c21-e7ad-4dc6-807a-c779ec7eecd5)

maybe something changed during the TS compilation and bundling?

```
$ node --version
v18.16.0
```

dolanmiu:
Weird. It does seem to return Buffer from the source code:

https://github.com/dolanmiu/docx/blob/master/src/export/packer/packer.ts#L31

And yes, in 8.2, I have moved the project from `webpack` to a more modern `vite` based build system, so maybe that's why?

tripodsan:
btw, my workaround:

```
  let buf = await Packer.toBuffer(doc);
  if (buf instanceof Uint8Array) {
    buf = Buffer.from(buf);
  }
  ```

**Comment by dolanmiu** [comment]
I've made a PR here to hopefully fix it:

https://github.com/dolanmiu/docx/pull/2244

**Comment by pooledge** [comment]
Just a heads up, presumably related. I had an edge case, where `bun` (v0.7.3 at moment of writing) stopped responding after repeatedly calling:
```ts
await Packer.toBuffer(document);
```
at least 3 times in a row. It came out during performance testing. I was able to overcome it by converting to `Buffer` on my own:
```ts
const blob = await Packer.toBlob(document);
const arrayBuffer = await blob.arrayBuffer();
const file = Buffer.from(arrayBuffer);
```

PR #2244, opened 2023-07-18
Fix `vite` build issues

dolanmiu:
Fixes https://github.com/dolanmiu/docx/issues/2222


Volper212:
That's a breaking change, isn't it?

dolanmiu:
No it's not

Tested it on:

Node
Regular html - iife
React
Angular

All fine

And the buffer change was intentional, it was `Uint8Array` before anyway, so now it is correct.

Issue #2237, opened 2023-07-14
Can't open documents generated with ^8.1.0 in Next.js API route using Packer.toBuffer()


rbourdon:
I use docx in a Next.js application (pages directory) in an api route with Packer.toBuffer(), and after v8.1.0 Word has failed to open the generated document.

Here's a simple reproduction on codesandbox: https://codesandbox.io/p/sandbox/exciting-meadow-k89f6g 

Reverting back to 8.0.4 resolves the issue.

Sliffcak:
Same Issue here.

joshkel:
I've encountered what I believe is the same issue. I have a [Hapi](https://hapi.dev/) backend that started returning unusable values when I upgraded to 8.1.0.

Since it started with 8.1.0, I believe it's the result of the migration to Vite in #2148.

From what I can tell:

The problem arises because the output of `Packet.toBuffer()` isn't a Node.js Buffer. It calls itself a Buffer (e.g., if you call JSON.stringify on it, "Buffer" appears in the output), and it's an instance of Uint8Array (just like a real Node.js Buffer), but it's not an instance of Buffer, so Hapi does not invoke its code path for serving a binary blob. I haven't looked into Next.js, but I'm guessing it's the same root problem.

From inspecting the `build/index.cjs` build of docx, it looks like docx -> jszip -> Buffer is resolving to the [third-party buffer package](https://github.com/feross/buffer) instead of using Node.js's native Buffer support. (In case it helps, my application is using the `build/index.cjs` build of docx.)

dolanmiu:
I upgraded the project from `webpack` to `vite` for better DevX and native ESM support

Yes, this is Vite doing some of it's own optimisations (I am guessing to make it compatiable for web and node etc).

Still looking into it

Sliffcak:
8.2.1 Fixed it for me as well.

dolanmiu:
Excellent, thanks for the testing everyone!

PR #2254, opened 2023-07-19
Draft: Unbundle dependencies

joshkel:
Instead of bundling all dependencies into docx's built JS (which is a common approach for browser environments), this lets them be imported through import / require (as is typical with NPM).  This reduces the size of docx's builds from ~711kB each to ~250kB each and avoids duplicating code for applications that are already using dependencies such as jszip.  It also avoids any Node.js polyfills, which is a mixed bag - that prevents problems such as #2237 but leaves it up to users to provide polyfills themselves if using a browser environment.

Code for listing dependencies is based on https://stackoverflow.com/a/74578129/25507

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -2,19 +2,12 @@ import { defineConfig } from "vitest/config";
 import { resolve } from "path";
 import tsconfigPaths from "vite-tsconfig-paths";
 import dts from "vite-plugin-dts";
-import { nodePolyfills } from "vite-plugin-node-polyfills";
+import pkg from './package.json';
 
 export default defineConfig({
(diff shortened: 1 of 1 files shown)


dolanmiu:
What would happen to the `umd` version of the library?

https://www.unpkg.com/docx@8.2.1/build/index.umd.js

It needs to be bundled for `umd` i think, so it can be used like so:

https://github.com/dolanmiu/docx/blob/master/demo/browser-demo.html

dolanmiu:
Yeah thought so:

image

Anyway to fix this you think?

dolanmiu:
Closing this

Issue #2082, opened 2023-04-18
Numbering instance restart number not work since v8.0.0 and work well in v7.8.2


hom:
In `v7.8.2`, the numbering of instance to restart number is work well, but since `v8.0.0`, the instance not work.

see:
```bash
# demo of 29
npm run demo
```

in `v7.8.2`:
<img width="577" alt="Screen Shot 2023-04-18 at 15 47 26" src="https://user-images.githubusercontent.com/23349958/232708328-f82cb38e-04f1-4d15-afe2-11ba6c070ec2.png">

in `v8.0.0`
<img width="468" alt="Screen Shot 2023-04-18 at 15 45 39" src="https://user-images.githubusercontent.com/23349958/232708039-7df39cda-764b-4f70-b673-1bd0ef8bc5b6.png">

Sliffcak:
@dolanmiu   have you found any work around? The issue is still present in latest v8.2.1

curlycabbage:
I am seeing the same issue

tripodsan:
me, too

PR #2338, opened 2023-09-13
fix: #2082 Numbering instance restart number not work since v8.0.0 an…

hom:
Hi, I found this bug #2082 some days ago and I fixed it now.

 With addination info, It caused at commit [29da523](https://github.com/dolanmiu/docx/commit/29da5235798283e55f67e81231a883ac05ba4b67). 

With this commit, I remove `@esbuild/win32-x64` dependency, the reason is  it causes`npm install` failed  in my macbook pro, and it is depended with `esbuild` but not this `docx` package.

The numbering restart count as normaly after I patched the `overrideLevel` option
It is the `29` demo result below :point_down:
image

I will fix as soon if this pr has any problem.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,6 @@
                 "xml-js": "^1.6.8"
             },
             "devDependencies": {
-                "@esbuild/win32-x64": "^0.18.3",
                 "@types/inquirer": "^9.0.3",
                 "@types/prompt": "^1.1.1",
                 "@types/unzipper": "^0.10.4",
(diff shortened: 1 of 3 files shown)


hom:
@dolanmiu Hi, can you take several minutes to review this pr? nice to receive you reply

dolanmiu:
This would be a breaking change, changing the API

hom:
I don't know why the workflow didn't runing normally too, do you know? The message tell me this pr need maintainer approve first...
[Demos](https://github.com/dolanmiu/docx/actions/workflows/demos.yml)

Issue #2293, opened 2023-08-09
How to make the content after patchDocument have the same style as the key?


KaneOne:
As shown in the screenshot below, I changed the font size of the paragraph, when using patchDocument, the replaced content does not use the font and style of the key but uses the default font instead. How can I set it so that the replaced content has the same style as the key? Thank you.
![image](https://github.com/dolanmiu/docx/assets/1954595/df6d781f-91e2-40fe-9fd3-0e81575d2e33)

![image](https://github.com/dolanmiu/docx/assets/1954595/3f253b9d-857c-4357-b97f-c1733ec5cad2)

KaneOne:
Hi, @dolanmiu 

May I ask if there is a plan to fix this bug? I hope to use the docx library in my project. Thank you.

KaneOne:
Working very well, thank you.

PR #2387, opened 2023-10-18
Fix style being missed on patchDocument

wilkmaia:
There were scenarios in which patching a document would result in loss of style for the template runs and, possibly, their right-adjacent run as well (post-splitting). That was due to the run style elements not being added to the newly created runs.

This commit addresses this issue by introducing a new, optional, flag for the `patchDocument` function: `keepOriginalStyles`. It defaults to `false` (current behavior) and, when `true`, ensures that there is no loss of styling.

This should address https://github.com/dolanmiu/docx/issues/2293. Fixes https://github.com/dolanmiu/docx/issues/2293.


dolanmiu:
Fixes https://github.com/dolanmiu/docx/issues/2293

PR #2397, opened 2023-10-26
fix: #2082 Numbering instance restart number not work since v8.0.0 and work well in v7.8.2

hom:
Hi, I found this bug #2082 some days ago and I fixed it now.

 With addination info, It caused at commit [29da523](https://github.com/dolanmiu/docx/commit/29da5235798283e55f67e81231a883ac05ba4b67). 

With this commit, I remove `@esbuild/win32-x64` dependency, the reason is  it causes`npm install` failed  in my macbook pro, and it is depended with `esbuild` but not this `docx` package.

The numbering restart count as normaly after I patched the `overrideLevel` option
It is the `29` demo result below :point_down:
image

I will fix as soon if this pr has any problem.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,6 @@
                 "xml-js": "^1.6.8"
             },
             "devDependencies": {
-                "@esbuild/win32-x64": "^0.18.3",
                 "@types/inquirer": "^9.0.3",
                 "@types/prompt": "^1.1.1",
                 "@types/unzipper": "^0.10.4",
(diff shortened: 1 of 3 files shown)


hom:
Why the workflow status is `Action required`

hom:
Please see this [workflows awaiting approval](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

hom:
> To help prevent this, workflows on pull requests to public repositories from some outside contributors will not run automatically, and might need to be approved first. By default, all first-time contributors require approval to run workflows.

**By default, all first-time contributors require approval to run workflows.**, It's my first time to pr this repository: `docx`

dolanmiu:
Running it now, works now

dolanmiu:
Thank you, merged

PR #2439, opened 2023-12-06
Run prop snapToGrid should be false

morlay:


morlay:
BTW

should snapToGrid be a Paragraph prop?

I found it under `w:pPr` from software created.

dolanmiu:
> should snapToGrid be a Paragraph prop?

From checking, its both actually

A paragraph prop and a run prop

morlay:
> From checking, its both actually
> 
> A paragraph prop and a run prop

seems missing it in Paragraph.
https://github.com/dolanmiu/docx/blob/master/src/file/paragraph/properties.ts

PR #2445, opened 2023-12-09
Change all project enums to objects with as const

Aemulatio:
Enums after compile looks so gross and they needs a little more space in comparison with objects. This PR should reduce final bundle size

Official documentation say that [we dont need to use **enum**s if we can use objects with **as const**](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums)

So that forced me to do this PR


dolanmiu:
Thanks for this. 
I think I agree with you. Someone at work said the exact same thing about our work projects

src/file/document/body/section-properties/properties/doc-grid.ts
```diff
@@ -17,14 +17,17 @@ import { decimalNumber } from "@util/values";
...
-    LINES_AND_CHARS = "linesAndChars",
-    SNAP_TO_CHARS = "snapToChars",
-}
+/* eslint-disable @typescript-eslint/naming-convention */
+export const DocumentGridType = {
+    DEFAULT: "default",
+    LINES: "lines",
+    LINES_AND_CHARS: "linesAndChars",
+    SNAP_TO_CHARS: "snapToChars",
+} as const;
```
hshoja, line 26: having this is good, but it would be hard to use types in code base, also using PascalCase for const not a good practice I would recommand to have this 

```suggestion
const documentGridType = {
    DEFAULT: "default",
    LINES: "lines",
    LINES_AND_CHARS: "linesAndChars",
    SNAP_TO_CHARS: "snapToChars",
} as const;

export type DocumentGridType = (typeof documentGridType)[keyof typeof documentGridType]
```

I can make a pr to fix it

Aemulatio: I've tried to keep naming of enums, otherwise projects after update package version will fall to errors

If talk about pascal, camel, etc. cases, I'd like to use upper case to show their constant state

dolanmiu: If we make a breaking change, then it has to be a v9 update

> I'd like to use upper case to show their constant state

Yes UPPER_CASE to show constants sounds nice to me

castortech: I just forked remark-docx which uses this project and was on version 8.2.0 and this was a breaking change. Thanks to @hshoja for providing a solution. I am no TS expert and never saw that spending hours trying to figure it out.

PR #2485, opened 2023-12-25
Page number in section

dolanmiu:


tungnat97:
Did you change "current page in section" to "current section"? I want to insert current page of section to footer but can't find the option.

PR #2523, opened 2024-01-08
Export FileChild class

YaoKaiLun:
Try to do something like below:

```
  const children: FileChild[] = [];

  for (const node of nodes) {
    if (Element.isElement(node)) {
      const child = convertComponent(node);
      children.push(child);
    }
  }

  const doc = new Document({
    sections: [
      {
        properties: {},
        children,
      },
    ],
  });
```

Now i have to do like below:

```
import { ISectionOptions } from 'docx';

const sectionOptionsInstance: ISectionOptions = {
  children: [],
};

type FileChild = typeof sectionOptionsInstance['children'][number];
```

### DIFF
--- a/src/file/index.ts
+++ b/src/file/index.ts
@@ -1,6 +1,7 @@
 export * from "./paragraph";
 export * from "./table";
 export * from "./file";
+export * from "./file-child";
 export * from "./numbering";
 export * from "./media";
 export * from "./drawing";


YaoKaiLun:
@dolanmiu please review my PR

YaoKaiLun:
@dolanmiu Could you publish a new npm version 🙏?

PR #2531, opened 2024-01-09
Allow disabling numbering inherited from a paragraph style

kalda341:


kalda341:
Hi @dolanmiu, have you had a chance to look over this?

src/file/paragraph/properties.ts
```diff
@@ -135,6 +137,8 @@ export class ParagraphProperties extends IgnoreIfEmptyXmlComponent {
             });
 
             this.push(new NumberProperties(`${options.numbering.reference}-${options.numbering.instance ?? 0}`, options.numbering.level));
+        } else if (options.numbering === false) {
```
dolanmiu, line 140: Better to check for undefined I think rather than false, so it is idiomatic with this library

kalda341: If we check for undefined then we'll be disabling inherited numbering by default (not what we're looking for), unless we treat undefined different to a missing value.
Maybe null is a better fit than undefined then?

kalda341: @dolanmiu Any thoughts?

dolanmiu: I see what you mean

Yes looks good, I like how it can be set to `false`, which has clear intent

Approved, merging

Issue #2680, opened 2024-05-05
Repeated images are written several times


mustache1up:
When using the same image in the document, each instance of the image results in a different file inside the generated docx file.

Documents with reused images gets linearly bigger.

Minimum example:

```javascript
import { Document, ImageRun, Packer, Paragraph } from "docx";

const imageBase64Data = "iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAFUlEQVR42mP8z8BQz0AEYBxVSF+FABJADveWkH6oAAAAAElFTkSuQmCC";

const doc = new Document({
    sections: [
        {
            children: [
                new Paragraph({
                    children: [
                        new ImageRun({
                            data: imageBase64Data, // image
                            transformation: {
                                width: 100,
                                height: 100,
                            },
                        }),
                    ],
                }),
                new Paragraph({
                    children: [
                        new ImageRun({
                            data: imageBase64Data, // same image
                            transformation: {
                                width: 200,
                                height: 200,
                            },
                        }),
                    ],
                }),
            ],
        },
    ],
});

Packer.toBuffer(doc).then((buffer) => {
    fs.writeFileSync("same_image_twice.docx", buffer); // two identical media files inside the docx
});
```

mustache1up:
We could use the git uses for a while now, using a digest of the file as [`uniqueId` of the image](https://github.com/dolanmiu/docx/blob/022b25cfcd33cd2ab9456f4b16cbc01ec3a1ed63/src/file/paragraph/run/image-run.ts#L86C37-L86C44). 

Git uses SHA1, I'll test if it work well in the ImageRun class.

mustache1up:
Seems to work great. Adding tests to spec files in order to open a PR.

Issue #849, opened 2021-03-23
Allow to specify image fileName and emu dimensions


tripodsan:
- allowing to specify the `fileName` would help reducing the document size if the same image is included several times
- allowing to specify the `transformation.emus` separately would allow to better set the image dimensions irrespective of the pixels.

tripodsan:
currently, I need to hack it like this:

```js
  const drawing = new Drawing(imageData, { floating: false });
  // hack to get document properties to set alt text
  if (node.title || node.alt) {
    const docProps = findXMLComponent(drawing, 'wp:inline/wp:docPr/_attr');
    if (docProps && docProps.root) {
      docProps.root.name = title || '';
      docProps.root.descr = alt;
    }
  }

  // create picture
  const pic = new ImageRun({
    data: data.buffer,
    transformation: data.dimensions,
  });
  // replace drawing with custom filename and emu dimensions
  const oldDrawing = findXMLComponent(pic, 'w:drawing');
  const idx = pic.root.indexOf(oldDrawing);
  if (idx >= 0) {
    pic.root.splice(idx, 1);
  }
  pic.root.push(drawing);
  pic.key = key;
  pic.imageData = imageData;
```

dolanmiu:
> allowing to specify the fileName would help reducing the document size if the same image is included several times

This one is done now!

> allowing to specify the transformation.emus separately would allow to better set the image dimensions irrespective of the pixels.

Hmm not so sure, `emu` is is not as intuitive, and is just a scale factor of 9525, so maybe it isn't worth the effort

Can raise again if people have interest

tripodsan:
thank you.

> Hmm not so sure, emu is is not as intuitive, and is just a scale factor of 9525, so maybe it isn't worth the effort

but you are already using `emu` for the `floating` offsets.

The problem I found is, that if I want to keep the information about the original dimensions of the image, but scale it differently, eg restricting the width/height of an image inside a table, I need to set the `emus`.

eg:

```js
  let x = data.dimensions.width * 9525;
  let y = data.dimensions.height * 9525;
  const limits = ctx.tableAlign ? LIMITS_TABLE : LIMITS;
  if (x > limits.width) {
    y = Math.round((limits.width * y) / x);
    x = limits.width;
  }
  if (y > limits.height) {
    x = Math.round((limits.height * x) / y);
    y = limits.height;
  }

  const imageData = {
    stream: data.buffer,
    fileName: data.key,
    transformation: {
      pixels: {
[...]

PR #2681, opened 2024-05-05
Reuse images: change ImageRun keys to be based on image data content

mustache1up:
In order to generate only one media file for repeated images with the same data the uniqueId of the image can be based on the image data SHA1 digest (same strategy git uses internally).

Solves #2680 and partially #849


dolanmiu:
This seems like a good idea

dolanmiu (approved):

PR #2698, opened 2024-07-03
fix: Ensure necessary namespaces are in patched doc

### LINKED ISSUE #2697: "Word experienced an error" from missing xmlns
We're using the [patcher](https://docx.js.org/#/usage/patcher) feature to generate Word documents from user-submitted templates. One of our templates has a very minimal `<w:document>` header within its `word/document.xml`:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
```

When we open the resulting Word document, Word reports an error:

> Word experienced an error trying to open the file.
> Try these suggestions.
> * Check the file permissions for the document or drive.
> * Make sure there is sufficient free memory and disk space.
> * Open the file with the Text Recovery converter.

xmllint seems to think that the problem is caused by missing xmlns namespaces:

```
word/document.xml:1: namespace error : Namespace prefix wp on inline is not defined
:r></w:p><w:p><w:r><w:drawing><wp:inline distT="0" distB="0" distL="0" distR="0"
                                                                               ^
word/document.xml:1: namespace error : Namespace prefix wp on extent is not defined
ine distT="0" distB="0" distL="0" distR="0"><wp:extent cx="6400800" cy="4800600"
                                                                               ^
word/document.xml:1: namespace error : Namespace prefix wp on effectExtent is not defined
"><wp:extent cx="6400800" cy="4800600"/><wp:effectExtent t="0" r="0" b="0" l="0"
                                                                               ^
```

If I manually add the missing namespaces, Word seems happy.

joshkel:
Word for the web, for example, may generate .docx files with a very limited set of XML namespaces:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">
```

Using one of these files as the template / source for a docx.js patch may result in errors from missing namespaces.

To fix this, I added the handful of namespaces that appear to be needed by docx.js's elements. It may be better to include the full set of namespaces from [the Document class](https://github.com/dolanmiu/docx/blob/8.5.0/src/file/document/document.ts#L39). If you'd prefer I do that, please let me know.

I did some refactoring to simplify this and to cut down on duplicate namespace references. If you'd prefer I handle this differently, please let me know. 

Fixes #2697


dolanmiu:
Seems good, need to make the CI to start. Unsure why it isn't starting

PR #2705, opened 2024-07-29
Slate to docx

pubuzhixing8:


dolanmiu:
I do not think this should be added as a dependency if it's only used in `demos`

dolanmiu:
Closing as `slate` is out of scope for this library. Feel free to make a new library and use `docx` as a dependency

Issue #2635, opened 2024-03-11
nodenext: Module '"docx"' has no exported member 'Paragraph'.


Spitzbua:
# Cannot import any other Types but document using Basic Usage Example

`
import { Document, Packer, Paragraph, TextRun } from "docx";
`

leads to following error: Module '"docx"' has no exported member 'Paragraph'

I tried several import variants, but none of them was working: 
https://github.com/dolanmiu/docx/issues/1201, https://github.com/dolanmiu/docx/issues/2365

Is it a misconfiguration or an library issue? Any help appreciated!

Options:

> {
  "compilerOptions": {
    "module": "nodenext",
    "moduleResolution": "nodenext",
    "noImplicitReturns": true,
    "noUnusedLocals": false,
    "skipLibCheck": true,
    "outDir": "lib",
    "sourceMap": true,
    "strict": true,
    "target": "es6",
    "typeRoots": [
      "./node_modules/@types",
      "./types"
    ]
  },
  "compileOnSave": true,
  "include": [
    "src"
  ]
}

OperationsOfficerPRMG:
I'm also experiencing this, is anyone able to please help? 

I was able to use this module in pure JS but when switching to TS it won't compile and gives many errors which look like this:

> Property 'Packer' does not exist on type 'typeof import("./node_modules/docx/build/index", { with: { "resolution-mode": "import" } })'.

The compiler will only seem to accept something like the following (non-sensical) code. 

>const { Document } = docx;
>const { Packer, Paragraph, Table, TableCell, TableRow, Header, HeadingLevel, Footer, LevelFormat, TextRun, WidthType, AlignmentType, VerticalAlign, convertMillimetersToTwip, TableBorders, ShadingType, BorderStyle, ImageRun, HorizontalPositionAlign, VerticalPositionAlign, TextWrappingType, UnderlineType, FrameAnchorType } = Document;

Which gives errors like 
> TypeError: convertMillimetersToTwip is not a function

Here's my tsconfig,json

> {
>   "$schema": "https://json.schemastore.org/tsconfig",
>   "display": "Node 21",
>   "_version": "21.0.0",
>   "files": [
>     "index.mts",
>     "index.d.ts"
>   ],
>   "compilerOptions": {
>     "lib": [
>       "es2023",
>       "es5",
>       "es6",
>       "dom",
>       "dom.iterable"
>     ],
>     "module": "node16",
>     "target": "es2022",
>     "strict": true,
[...]

PR #2714, opened 2024-08-20
Enable rollupTypes in the vite build configuration.

sflanker:
This resolves issues with in dependent TypeScipt packages that are using moduleResolution: NodeNext Modern ES modules cannot use extensionless relative paths in imports.

Fixes https://github.com/dolanmiu/docx/issues/2635


dolanmiu:
Thank you for this, merging

Issue #2716, opened 2024-08-20
`replacer` duplicates line breaks when `keepOriginalStyles: true`


SebKranz:
I presume that the following line duplicates non-textual elements in order to move `w:Pr` tags over to the inserted, and the right run.

https://github.com/dolanmiu/docx/blob/962795743c4f21d84c52e47e4f82dcf4ca536dc3/src/patcher/replacer.ts#L68-L82

However, the function for filtering non-textual elements currently doesn't account for `<w:br>`. 

As a result, the `w:br` tag is added to the inserted run and also the right.

SebKranz:
I've added a PR that solves this for `w:br` tags.

However, looking at section 17.3.3 of the Office Open XML spec at https://ecma-international.org/publications-and-standards/standards/ecma-376/, there are many tags that can exist within a text run.
In fact, it looks to me like `w:Pr` is the **_only_** that should be duplicated.

PR #2717, opened 2024-08-21
Fix duplicated br tags

SebKranz:
This PR demonstrates the issue described in https://github.com/dolanmiu/docx/issues/2716 and a possible fix.

I'm marking this as draft, because I would like check a few more details:

- are other tags affected (eg. `w:tab`)
- I should probably create a new test rather than modifying the existing one

Fixes: https://github.com/dolanmiu/docx/issues/2716


dolanmiu:
Thank you, we came to the same conclusion

dolanmiu:
You're right, tabs are also affected, I will fix this too

SebKranz:
Based on the list of allowed children at http://officeopenxml.com/WPtext.php, I think we should change this to only include `w:Pr`. 

`w:Pr` is the only tag related to styling and all others represent visual elements that shouldn't be duplicated.

Issue #2855, opened 2024-11-25
patchDocument: incorrect output when patches have no text


SebKranz:
I encountered weird results when using the document replacer:

- the `SPLIT_TOKEN` (ɵ) appeared in the final document.
- Placeholders within the same paragraph appeared out of order.

The following seems to be happening:

1. I was instantiating a TextRun with an empty-string
    ```ts
    {
      type: PatchType.PARAGRAPH,
      children: new TextRun({ text: "" }),
    }
    ```
2. since I was calling the constructor with an object, rather than a plain string (`new TextRun("")`), the inner `Text` was created with `space: "default"`, not `space: "preserve"`
3. the formatter drops the empty `w:t`.
4. the replacer inserts the `SPLIT_TOKEN`, and then crashes when trying to access the now missing `elements` property here: https://github.com/dolanmiu/docx/blob/c6bb2556419f49f0ea42e774a20f5fa5f7450350/src/patcher/replacer.ts#L70C1-L74C1
5. because the outer functions catches all errors, the merge continues and future runs on the same paragraph use the incorrect split-token.

I think there are three problems to be addressed:

- the code for detecting when all patches are found is masking all unintended errors. This should be communicated in the returned value, not an error.
- the `replacer` should handle empty Runs.
- there should be an option for controlling the white-space `TextRun({ space: "preserve", text: "..." })`

This issue does not occur when:

- `keepOriginalStyles` is off
- we instantiate the Run with a string argument (`new TextRun("...")`), since that will set `space: "preserve"`

PR #2856, opened 2024-11-25
fix(replacer): errors suppressed by catch statement

SebKranz:
This PR changes the signature of `replacer` so that it returns a boolean to signify whether patches were found rather than throwing.

Impact on clients:

- The replacer isn't exported. Therefore it should be safe to change the signature
- Users who have previously had hidden errors in their projects might now encounter crashes. 

Relates to https://github.com/dolanmiu/docx/issues/2855


dolanmiu (approved):

dolanmiu:
Does this fix https://github.com/dolanmiu/docx/issues/2855?

PR #2875, opened 2024-12-02
Fix replacer error on patches with empty Runs

SebKranz:
This PR fixes the error in #2855, which occurs because of two minor issues:

1. `TextRun("")` and `TextRun({ text: "" })` counterintuitively didn't do the same thing. One creates an inner `Text` and one doesn't. This PR changes `TextRun("")`, so that both always create the inner Text.
2. `replacer` didn't handle patches with empty `Run`s. This is also fixed by this PR.


dolanmiu:
Thank you!

Issue #2667, opened 2024-04-04
Using embedded fonts for client side generated files


xong:
Hi,

I am trying to embed custom fonts.
The docx-file ist generated at the client browser.

Please consider following example:
```javascript
// create document0
import {Document, Packer} from "docx";

// base64 encoded demo/assets/Pacifico.ttf (Insert by yourself since it would burst this comment.)
const font = atob('...');

const doc = new Document({
    sections: [],
    fonts: [{ name: 'Pacifico', data: font }], // this is the error causing line
});

Packer.toBlob(doc).then((blob) => {
    let a = document.createElement('a');
    a.href = window.URL.createObjectURL(blob);
    a.download = 'example.docx';
    a.click();
});
```

I get following error message for `index.mjs:18833`:
```
Uncaught (in promise) TypeError: bytesToObfuscate.map is not a function
```

Is there any way to embed custom fonts for client side use?

Your help will be very much appreciated.

Robert

PR #2898, opened 2024-12-11
Fix: custom fonts support on the Browser

Dabolus:
Right now, it's not possible to embed custom fonts when generating a docx from a Browser without polyfilling Buffer because the obfuscator uses native Node.js Buffers without checking for their availability.

This PR makes the obfuscator use Uint8Arrays instead of Buffers, so that it can also work in the Browser without polyfilling.

I also renamed the related tests file because there was a small typo.

Resolves: https://github.com/dolanmiu/docx/issues/2667


xong:
Hi @Dabolus,

can you provide an example with which your able to embed custom fonts?

Kind regards,
Robert

xong:
> can you provide an example with which you are able to embed custom fonts?

Example can be found here: https://github.com/dolanmiu/docx/issues/2667#issuecomment-2647651008

PR #2937, opened 2025-01-16
hotfix: AlignmentType Justified typo

acn-meljosephfilio:
image


dolanmiu:
It is correct before, no need for this, but thank you anyway!

Docs:

https://c-rex.net/samples/ooxml/e1/Part4/OOXML_P4_DOCX_ST_Jc_topic_ID0EXZZ2.html

Both is justified

acn-meljosephfilio:
Apologies for that! It doesn't seem to work in my implementation, but it seems to be working correctly now. It might just be my misunderstanding of my code. My bad!

PR #2941, opened 2025-01-22
Feat: subfile overrides

### LINKED ISSUE #2938: Extending zip output with more sub-files
Hi @dolanmiu  Thanks for this great library.

I'm just running into an issue with trying to add a `commentsExtended.xml` subfile to the exported zip file (for MS Word compatibility). 

From what I can tell, the simplest way to make this possible would be to publicly expose the `Compiler` class, which is not currently part of the public api. Alternatively the `compiler` field on the `Packer` class could be made public.

In either case, it would enable user code to call `Compiler.compile` directly, get the returned `JSZip`, ad additional files to the `JSZip`, and finally generate the zip.

If you're open to the idea, I'm happy to open a PR to make the changes. Let me know what you think

adamSherwoodGenieAI:
closes #2938 

* adds an `overrides` optional parameter to `Compiler.compile` and the `Packer` methods
* adds tests for the `overrides` parameter
* updates `Packer` usage docs
  * adds usage example for `Packer.toString` and `Packer.toStream` which were undocumented
  * adds brief explanation for the optional `prettify` parameter which was undocumented
  * adds usage example for `overrides` parameter

Let me know if this needs anything else

### DIFF
--- a/docs/usage/packers.md
+++ b/docs/usage/packers.md
@@ -2,7 +2,7 @@
 
 > Packers are the way in which `docx` turns your code into `.docx` format. It is completely decoupled from the `docx.Document`.
 
-Packers works in both a node and browser environment (Angular etc). Now, the packer returns a `Buffer`, `Blob` or `base64 string`. It is up to you to take that and persist it with node's `fs`, send it down as a downloadable file, or anything else you wish. As of `version 4+`, this library will not have options to export to PDF.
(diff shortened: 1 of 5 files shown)


adamSherwoodGenieAI:
@dolanmiu This is the PR related to that conversation we had in #2938

dolanmiu:
Looks good, merged

PR #2943, opened 2025-01-24
fix: patchDocument, looks for namspaces more carefully over whole doc…

13W:
…ument

added support of different placeholder brackets, default `{{`: `}}`

extended inputDataType to accept JSZip

### DIFF
--- a/src/patcher/from-docx.spec.ts
+++ b/src/patcher/from-docx.spec.ts
@@ -202,15 +202,11 @@ describe("from-docx", () => {
     describe("patchDocument", () => {
         describe("document.xml and [Content_Types].xml", () => {
             beforeEach(() => {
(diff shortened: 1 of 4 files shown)


dolanmiu:
Can you add a test to your PR?

13W:
sorry guys, i don't think that i have enough time to write tests... and by specific of my work - i have to fix few bugs in your library to satisfy our requirements...
it would be good if someone fix the problems that i faced within your workflow...
i can explain each fix that i did if you wish

dolanmiu:
Hi @13W

Sure, yes, can you explain each fix? Maybe it can be a comment in the code

dolanmiu:
Almost ready, just need to write tests

PR #3041, opened 2025-04-15
feat: Added NumberedItemReference

Alexnortung:
This PR adds a `NumberedItemReference` class.

This makes it simpler to make cross reference to a numbered item within the document.

One caveat is that you have to provide a cached value, just as you would with a `SimpleFIeld` I am not sure if there is a way to automate this at the moment.

If you see anything I can do better, please let me know.


dolanmiu:
Raised a new PR for it to fix linting and wrote tests:

https://github.com/dolanmiu/docx/pull/3042

PR #3042, opened 2025-04-16
feat: Added NumberedItemReference

dolanmiu:
Resolves: #3041

Adding linting and more tests to make it pass CI

Thank you!


Alexnortung:
Hi @dolanmiu,

Thanks for merging my commit! And thanks for the great work. I noticed that the lint was failing, saying 

>Use a `type` instead of an `interface`

However the documentation says ([link](https://docx.js.org/#/contribution-guidelines?id=interfaces-over-type-alias))

> Do not use type, but rather use Interfaces. type cannot be extended, and a class cannot implement it.

Maybe this should be changed? :)

dolanmiu:
Documentation updated. Thanks!

PR #3140, opened 2025-06-10
fix: canonicalize <w:rPr> element order for Word compatibility

helloendoftheworld:
### Problem

The ECMA-376 specification (5th edition) defines the `<w:rPr>` element (run properties) using an `<xsd:group>` with `<xsd:choice>`, which formally indicates that the order of child elements is not semantically significant. However, both tooling and rendering engines in the wild violate this assumption.

Two classes of breakage were observed:

1. **Validation failure**

The [OOXML Validator for VS Code](https://marketplace.visualstudio.com/items/?itemName=mikeebowen.ooxml-validator-vscode) (validated against `Microsoft 365 Subscription`) rejects otherwise valid DOCX files when `<w:rPr>` elements appear in non-canonical order. For example, when `<w:color>` follows `<w:u>`, the validator emits the following error:

```
Id: Sch_UnexpectedElementContentExpectingComplex
Description: The element has unexpected child element 'http://schemas.openxmlformats.org/wordprocessingml/2006/main:color'.
XPath: /w:document[1]/w:body[1]/w:p[1]/w:r[1]/w:rPr[1]
Part URI: /word/document.xml
NamespacesDefinitions: undefined
```

relevant code:
```
 <w:r>
    <w:rPr>
      <w:u w:val="single"/>
      <w:color w:val="d32f2f"/>
      <w:shd w:fill="388e3c" w:color="auto" w:val="solid"/>
    </w:rPr>
    <w:t xml:space="preserve">Hello guys</w:t>
</w:r>
```

This indicates that the validator treats the content model of `<w:rPr>` as **ordered**, even though the specification explicitly permits any one of the defined elements to appear, in any order, and at most once.

2. **Rendering issues in Word Online**

Empirical testing shows that MS Word Online silently drops or ignores formatting if the order of elements in `<w:rPr>` deviates from its internal expectations. I tested specifically:

* If `<w:strike>` appears after `<w:sz>` or `<w:szCs>`, strike-through formatting is ignored.
* If `<w:color>` appears after `<w:u>` (underline), the text color is ignored 

These failures are silent - Word Online does not emit an error or warning, but formatting is broken.

### fix

This patch reorders the construction logic in to match the canonical order.


dolanmiu:
Very interesting and good spot

helloendoftheworld:
Just fixed the tests.

PR #3272, opened 2025-10-01
fix: Allow zero as first level start number

giuscris:
This PR updates a check on `firstLevelStartNumber` inside `numbering.ts` to allow lists start with a 0 (zero) item if explicitly provided. Otherwise the logic is unchanged.

Before this PR even if you wanted a 0 item by explicitly setting 0 to the `start` property, it would become 1 as the first part of the conditional was evaluated as false.

I've checked and all tests pass, so I think the impact is minimal.


edoardocavazza:
Hello @dolanmiu, sorry for the mention, but is there any hope of getting this PR merged?

dolanmiu:
All merged! Thank you @edoardocavazza

PR #3311, opened 2025-11-06
fix: Multiple splits in a single run

FuegoFro:
There was an issue where the way split index was the last item being split, which broken when we separated items into left/right, leading to overwritten values or "token not found" errors.

This stops us searching the run as soon as we find a place to split, and also guarantees that we don't try to start looking for a split past the end of the paragraph.

I've added a test that reproduces the issue, and then fixed the implementation such that the test passes.

### DIFF
--- a/src/patcher/paragraph-split-inject.spec.ts
+++ b/src/patcher/paragraph-split-inject.spec.ts
@@ -276,6 +276,65 @@ describe("paragraph-split-inject", () => {
             });
         });
 
+        it("should put all content on the right when token is not found", () => {
+            const output = splitRunElement(
+                {
+                    name: "w:r",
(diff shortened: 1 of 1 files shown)


dolanmiu:
@FuegoFro Thanks for this! Could you look into this failing test?

FuegoFro:
Ah sorry about the delay here! I went ahead and updated the test to match the new behavior. That being said, that does mean there's a slight change in output when not finding a split, and I don't know if anyone might be relying on the existing behavior. If so, I can also add a quick special case to put everything on the left instead of the right when no split is found!

dolanmiu:
Thank you @FuegoFro! Merging

PR #3317, opened 2025-12-08
 Add table look support for conditional table formatting

paradoxloop:
Implement `<w:tblLook> `element to control which conditional formatting from table styles is applied.

 Table styles can define special formatting for first/last rows/columns and horizontal/vertical banding. The new `tableLook` option allows users to toggle which of these formatting options are applied table by table.

  Supported attributes:
  - firstRow
  - lastRow
  - firstColumn
  - lastColumn
  - noHBand; and
  - noVBand

  Changes:
  - Add TableLook component with ITableLookOptions type
  - Integrate tableLook option into Table component; and
  - Update documentation

  Reference: http://officeopenxml.com/WPtblLook.php

### DIFF
--- a/docs/usage/tables.md
+++ b/docs/usage/tables.md
@@ -354,12 +354,12 @@ Control which conditional formatting from a table style is applied. Table styles
 const table = new Table({
     style: "GridTable5Dark-Accent3",
     tableLook: {
-        firstRow: true,      // Apply first row formatting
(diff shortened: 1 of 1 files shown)


paradoxloop:
Screenshot 2025-12-08 at 4 32 31 pm
What is being controlled is the toggle on left of screenshot.

PR #3325, opened 2026-01-20
feat(footnotes): Support lists and images in footnotes

### LINKED ISSUE #2876: Adding lists to footnotes results in broken DOCX generation
Library generates broken DOCX when adding lists to footnotes.

**Example of issue reproduction:** https://codesandbox.io/p/sandbox/2sqj2q

After investigation, it seems to be a problem in the [https://github.com/dolanmiu/docx/blob/master/src/export/packer/next-compiler.ts](url). There is no logic that replaces the numbering for footnotes, as is done in other parts of DOCX.

As I can see, there is also no logic to image replacement for footnotes, so adding images to them will also break the file.

james-atticus:
Fixes #2876

Adds support for images and lists in footnotes:

image

Drive-by fix for lists in headers/footers:

image

### DIFF
--- a/docs/usage/footnotes.md
+++ b/docs/usage/footnotes.md
@@ -55,6 +55,101 @@ Insert `FootnoteReferenceRun` in paragraphs to create reference markers:
 - Automatically styled as superscript
 - References the corresponding footnote content
 
+## Images in Footnotes
+
+Footnotes can contain images using `ImageRun`:
+
+```ts
+import { Document, FootnoteReferenceRun, ImageRun, Packer, Paragraph, TextRun } from "docx";
+import * as fs from "fs";
+
+const doc = new Document({
+    footnotes: {
+        1: {
+            children: [
+                new Paragraph({
+                    children: [
+                        new ImageRun({
+                            type: "jpg",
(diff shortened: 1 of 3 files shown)


src/file/paragraph/properties.ts
```diff
@@ -221,7 +221,7 @@ export class ParagraphProperties extends IgnoreIfEmptyXmlComponent {
     }
 
     public prepForXml(context: IContext): IXmlableObject | undefined {
-        if (context.viewWrapper instanceof DocumentWrapper) {
+        if (!(context.viewWrapper instanceof FontWrapper)) {
```
james-atticus: There was already logic in `next-compiler.ts` to call `this.numberingReplacer.replace` for headers and footers, but the replacements weren't being put into `Numbering` because of this `viewWrapper` check. I had a look at what implements `IViewWrapper` and it looks like the only one we now want to exclude here is `FontWrapper`.

james-atticus:
@dolanmiu any chance you can take a look at this one?

dolanmiu:
@james-atticus Looks good, merging!

dolanmiu (approved):

PR #3328, opened 2026-01-28
Extend Table of Contents creation

volodymyroliinykv:
**Problem** 
I can't export TOC without popup (about update fields) and headings with custom styles 

**Solution**

1. Added possibility create TOC with children
2. Added dirty flag for begin element

### DIFF
--- a/src/file/table-of-contents/table-of-contents.spec.ts
+++ b/src/file/table-of-contents/table-of-contents.spec.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "vitest";
 
 import { Formatter } from "@export/formatter";
+import { Paragraph } from "@file/paragraph";
 
 import { TableOfContents } from "./table-of-contents";
 import { StyleLevel } from "./table-of-contents-properties";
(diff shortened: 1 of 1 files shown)


src/file/table-of-contents/table-of-contents.ts
```diff
@@ -4,14 +4,20 @@ import { FileChild } from "@file/file-child";
...
 import { ITableOfContentsOptions } from "./table-of-contents-properties";
 
 export class TableOfContents extends FileChild {
-    public constructor(alias: string = "Table of Contents", properties?: ITableOfContentsOptions) {
+    public constructor(
+        alias: string = "Table of Contents",
+        properties?: ITableOfContentsOptions,
+        contentChildren?: readonly (XmlComponent | string)[],
+        beginDirty: boolean = true,
+    ) {
```
12joan: Perhaps these additional options should be folded into `ITableOfContentsOptions` to avoid too many positional arguments?

I can see that we're currently passing these options as `properties` to the `FieldInstruction` constructor, but that could be addressed by doing something like this:
```ts
{ contentChildren, beginDirty, ...properties }: ITableOfContentsOptions = {}
```

dolanmiu:
Thank you for the PR, looks good. Merging

PR #3329, opened 2026-01-29
fix: pre-encode XML to UTF-8 to avoid surrogate pair corruption in JSZip

Yuof:
## Problem

When JSZip processes large XML content (> 16KB), it chunks strings using \substring()\ which operates on UTF-16 code units. This can split surrogate pairs for characters above U+FFFF (like emoji or Material Design Icons).

Each surrogate then gets encoded as a separate 3-byte UTF-8 sequence, producing invalid CESU-8 instead of proper UTF-8. This corrupts the docx file and causes XML parsing errors.

## Root Cause

JSZip's \DataWorker\ uses:
\\\javascript
DEFAULT_BLOCK_SIZE = 16 * 1024
data.substring(index, nextIndex)
\\\

The \Utf8EncodeWorker\ then processes each chunk independently, without handling split surrogates. I've opened a fix PR to JSZip: https://github.com/Stuk/jszip/pull/963

## Solution

This PR works around the issue by pre-encoding strings to UTF-8 using \TextEncoder\ before passing to \zip.file()\. Since we pass \Uint8Array\ instead of strings, JSZip skips the problematic \Utf8EncodeWorker\ entirely.

## Changes

- Added \ncodeUtf8()\ helper function in \
ext-compiler.ts\ and \rom-docx.ts\
- Modified \zip.file()\ calls to use pre-encoded UTF-8 bytes for XML content

## Testing

- Build succeeds
- TypeScript compiles without errors
- This is a minimal, low-risk change that ensures UTF-8 encoding happens correctly before JSZip processing

### DIFF
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -1,24 +1,19 @@
 - Getting Started
-
     - [Home](/)
     - [Quickstart Guide](quickstart.md)
 
 - Examples
-
     - [Demos](https://github.com/dolanmiu/docx/tree/master/demo)
 
 - API
-
     - [Documentation](https://docx.js.org/api/)
 
 - Core Concepts
-
     - [Document](usage/document.md)
     - [Sections](usage/sections.md)
     - [Convenience Functions](usage/convenience-functions.md)
 
 - Content Elements
-
     - [Paragraph](usage/paragraph.md)
         - [Text Frames](usage/text-frames.md)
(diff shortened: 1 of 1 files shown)


dolanmiu:
Looks good. I fixed tests and removed that TextEncoder fallback. All major browsers have this already, so it's ok.

Removed Buffer dependency, there was a PR which removed Buffer, so best to adhere to that

PR #3344, opened 2026-02-13
Merge external styles with default styles

### LINKED ISSUE #2702: "Styling with XML" Heading definition not taking effect
I need to define styles using XML. So, I tested this example code (demo/13-xml-styles.ts), but the example didn't execute as expected: line 15 of the code didn't take effect, it only generated the body format.

aghster:
Fixes #2702

### DIFF
--- a/src/file/file.ts
+++ b/src/file/file.ts
@@ -122,8 +122,17 @@ export class File {
         this.media = new Media();
 
         if (options.externalStyles !== undefined) {
-            const stylesFactory = new ExternalStylesFactory();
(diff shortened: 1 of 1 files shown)


aghster:
This is more complex than I thought, since `ExternalStylesFactory.newInstance()` returns a `Styles` class instance, whereas `DefaultStylesFactory.newInstance()` returns an `IStylesOptions` object, and only `IStylesOptions` has the `importedStyles` property. Closing, as ATM I have no time to dig deeper.

Issue #3120, opened 2025-05-26
Custom font works in Word but not in LibreOffice


ivmarcos:
I’m using a custom font following the example from the [docs](https://github.com/dolanmiu/docx/blob/495f2e48887e9a8f4f4bd2e0512c22b2bae86ae7/demo/91-custom-fonts.ts) and it shows up fine in Word, but LibreOffice (v25.x) doesn’t seem to pick it up. 

MS Word (correct):
![Image](https://github.com/user-attachments/assets/d940b29d-7cc2-49d2-8ead-766bd54aecb5)

LibreOffice Writer:
![Image](https://github.com/user-attachments/assets/4f4b69a2-0ef2-48b0-9a77-f94275ea3130)

At first I thought LibreOffice just didn’t support embedded fonts in DOCX, but I tested another .docx file created elsewhere using a embedded font, and it worked, so that’s not the issue.  As reference, this file contains the embedded font  "Ubuntu" which display correctly in Word and LibreOffice: https://calibre-ebook.com/downloads/demos/demo.docx.

Maybe something about how the font is defined in the generated file isn’t playing well with LibreOffice.

Issue #2521, opened 2024-01-05
Couple issues with embedded fonts


mattidupre:
I might be jumping the gun but I wanted to play with embedded fonts. I've noticed two issues.

Using Word 365 / MacOS:

1: Font names with spaces corrupt the docx file.
2. Every font I tried aside from Pacifico (which I had already installed) is failing to render. The document loads fine but the font disappears from the Word UI selection once it gets changed. I do see those fonts showing up in fontTable.xml file. Confirmed the fonts have no embedding restrictions. Repo using demo 91 and swapping for [Rubik](https://fonts.google.com/specimen/Rubik+Doodle+Shadow) or [Salsa](https://fonts.google.com/specimen/Salsa).

(I did notice that Word will retain a previously-installed font after it has been removed from the system.)

jcyovera:
I have the same trouble using ` "docx": "^8.5.0",` 
Scenario: I embed fonts on the code, `fonts: [{ name: "Pacifico," data: font, characterSet: CharacterSet.ANSI }],` but when I try to open the Word file, the Word app shows an alert that the file is corrupted.

The workaround is to embed the fonts and put a name without spaces, that works for me.

jcyovera:
You are right. The fonts are not loading on the document. It is an issue with the docx library because I follow the example in the repository.

vm11elim2:
i have same issue.... embedded fonts  only works for Pacifico....

vatsenko:
On my side - Corrupted file when space persist in font name (Microsoft Word error - Word found unreadable content ... )

YousefED:
I'm running into this as well. Could it be because there's no link from document.xml.rels to the fonttable? When I export a document with embedded font from Word, I see the following entry in `document.xml.rels`:

```xml
<Relationship Id="rId4" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/fontTable" Target="fontTable.xml"/>```

However, when I export a document with an embedded font using docx.js, this seems to be missing

PR #3428, opened 2026-04-30
fix: use sequential filenames for embedded fonts

### LINKED ISSUE #3019: Custom fonts with spaces in them corrupt MS word file
There was a similar issue from the past 

#2521 which reported this issue alongside another issue relating to `docx` not adding the fontTable.xml to the document.rels.xml 

It seems that only the latter issue was fixed, and this original issue with spaces in the filename remains.

I believe it should be a simple fix, if you look at how MS word embeds custom fonts, it just names each font in the fonts/ dir as font1.odttf etc.

## Workaround
I am having a look into this issue, as it seems that custom fonts with spaces do not work at all - using the workaround mentioned in this other issue replacing spaces with `-` does not seem to work. 

The file is no longer corrupted, but the font families do not match up with font table properly in this intance

ctjlewis:
## Summary

Embedded fonts with spaces (or other non-ASCII characters) in their family name caused Word to show its "Word found unreadable content" recovery prompt on open. The package zip entries were named after the user-facing family (e.g. `word/fonts/EB Garamond.odttf`), and Word treats those paths as literal filenames that must be plain.

This PR switches to sequential `fonts/font1.odttf`, `fonts/font2.odttf`, … in both the relationship `Target` and the zip entry. The user-facing family name continues to live only in `<w:font name="...">`, so no API change.

Closes #3019.

## Context

Refs #2521 — that issue reported two problems together. PR #2800 fixed one of them (the missing `Relationship` from `document.xml.rels` to `fontTable.xml`). The other (spaces in the package filename) was left unfixed and resurfaced as #3019. This PR completes the pair. May also help #3120 — LibreOffice silently dropping embedded custom fonts.

## What changed

- `FontWrapper`: relationship `Target`s are now `fonts/font<N>.odttf`.
- `next-compiler`: zip entries match the same sequential names.
- The user-facing family name lives only in `<w:font name="...">` — unchanged.

## Tests

- New `font-wrapper.spec.ts` — verifies relationship Targets are sequential.
- New case in `next-compiler.spec.ts` — verifies zip entries are `font1.odttf` etc., with no family-name paths leaking through.

## Test plan

- [x] Existing font tests pass (`vitest run src/file/fonts src/export/packer/next-compiler`)
- [x] Generated docx with `EB Garamond` (space in family name) opens in Word with no recovery prompt
- [x] New regression tests cover both the relationship Target and the zip entry path

### DIFF
--- a/src/export/packer/next-compiler.spec.ts
+++ b/src/export/packer/next-compiler.spec.ts
@@ -253,5 +253,28 @@ describe("Compiler", () => {
 
             compiler.compile(file);
         });
+
+        it("should write embedded fonts to sequential filenames in the zip (no spaces or special chars from family name)", () => {
+            // Regression for https://github.com/dolanmiu/docx/issues/3019 —
+            // fonts whose user-facing family name contains spaces / non-ASCII
+            // used to be written into the package zip with the family name as
+            // the filename (e.g. `EB Garamond.odttf`). Word rejects those
+            // paths and shows a "found unreadable content" recovery prompt on
+            // open. Sequential names side-step that.
+            const file = new File({
+                sections: [],
+                fonts: [
+                    { name: "EB Garamond", data: Buffer.from("") },
+                    { name: "Source Serif 4", data: Buffer.from("") },
+                ],
+            });
+
+            const zip = compiler.compile(file);
+            const fileNames = Object.keys(zip.files);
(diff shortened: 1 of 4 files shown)


dolanmiu:
Awesome! Will merge

PR #3429, opened 2026-05-04
fix: `paragraphStyles` should not include `w:pStyle`

Alexnortung:
Currently when `paragraphStyles` has a style with a numbering, a `w:pStyle` tag with `ListParagraph` style is inserted into the paragraph style. This causes weird behavior in microsoft word, where choosing one of the custom styles with numbering always make it fall back to `ListParagraph`.

I am not sure how to fix this in a good way, so I have just added some tests that someone else (or myself if I get the time) can use as a starting point.

### DIFF
--- a/src/file/styles/style/paragraph-style.spec.ts
+++ b/src/file/styles/style/paragraph-style.spec.ts
@@ -106,6 +106,43 @@ describe("ParagraphStyle", () => {
             });
         });
 
+        it("should not include w:pStyle for paragraph styles with numbering", () => {
+            const style = new StyleForParagraph({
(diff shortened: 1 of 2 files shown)


dolanmiu:
Thanks for the report and the tests, @Alexnortung. The fork is organisation-owned, so I couldn't push to this branch. I picked the work up in https://github.com/dolanmiu/docx/pull/3519, which keeps your commit, adds the fix and the remaining tests, and supersedes this PR.

dolanmiu:
Superceeded by https://github.com/dolanmiu/docx/pull/3519

PR #3432, opened 2026-05-08
feat: add track change support to images

james-atticus:
Building on top of the approach from https://github.com/dolanmiu/docx/pull/3326, this adds support for tracked insert/delete of image runs. I considered doing something more similar to `Inserted/DeletedTextRun` but this feels like a more ergonomic API.


dolanmiu (approved):

dolanmiu:
Thank you! Merged

PR #3472, opened 2026-06-23
feat: emit word/commentsIds.xml with durable comment ids (w16cid)

alexvcasillas:
## Summary

Adds support for the `word/commentsIds.xml` OOXML part, which maps each comment's volatile paragraph id (`w14:paraId`) to a stable `w16cid:durableId` that Word preserves across edits (unlike the volatile `w:id` in `comments.xml`).

This mirrors the existing `commentsExtended.xml` feature end to end. A new optional `durableId` field on a comment opts a document into the part: when any comment carries a `durableId`, the library now generates and registers `commentsIds.xml`.

## Why

Word, and the apps and services that interoperate with it (for example collaborative editors that round-trip DOCX), rely on `w16cid:durableId` to track comment identity stably across save and edit cycles. Without it, comment ids can be reassigned and external systems lose the ability to reconcile a comment with its original. `docx` already emits `comments.xml` and `commentsExtended.xml`, but had no way to write the durable-id part.

### Public API

- `ICommentOptions` gains an optional `readonly durableId?: string`. When it is omitted on a comment that lives in a `commentsIds`-bearing document, the comment's generated `paraId` is reused as its `durableId` (matching how Word emits the part).
- New exported `CommentsIds` component (sibling of `CommentsExtended`).
- New exported `ICommentIdData` type (sibling of `ICommentThreadData`).
- New `File#CommentsIds` getter (sibling of `File#CommentsExtended`).

### OOXML parts produced

- Root element `w16cid:commentsIds` with namespaces:
  - `xmlns:w16cid="http://schemas.microsoft.com/office/word/2016/wordml/cid"`
  - `xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"`
  - `mc:Ignorable="w16cid"`
- Child `w16cid:commentId` elements with `w16cid:paraId` and `w16cid:durableId`.
- Content-type override `application/vnd.openxmlformats-officedocument.wordprocessingml.commentsIds+xml` for `/word/commentsIds.xml`.
- Relationship type `http://schemas.microsoft.com/office/2016/09/relationships/commentsIds` pointing at `commentsIds.xml`.

### Generation logic

- `Comments` now assigns deterministic `paraId`s when threading is active or when any comment carries a `durableId` (previously threading only). It exposes a new `CommentIdsData` getter (sibling of `ThreadData`), built only when a `durableId` is present.
- `File` wires up the `CommentsIds` part, relationship, and content-type override whenever `Comments#CommentIdsData` is set (sibling of the `commentsExtended` wiring).
- The packer and compiler serialize the part to `word/commentsIds.xml` (sibling of the existing `commentsExtended.xml` block).

## Files touched

- `src/file/paragraph/run/comment-run.ts`: `durableId` option, `ICommentIdData`, generator and getter
- `src/file/paragraph/run/comments-extended.ts`: `CommentsIds` component
- `src/file/file.ts`: part wiring, relationship, content-type, getter
- `src/file/content-types/content-types.ts`: `addCommentsIds()`
- `src/file/relationships/relationship/relationship.ts`: new relationship type
- `src/export/packer/next-compiler.ts`: serialize part to `word/commentsIds.xml`

## Tests

Added to the existing specs (vitest), covering threaded and single (non-threaded) cases:

- `comment-run.spec.ts`: `CommentIdsData` mapping, paraId fallback, threaded plus durable, paraId injection
- `comments-extended.spec.ts`: `CommentsIds` XML serialization
- `file.spec.ts`: `File#CommentsIds` presence and absence
- `next-compiler.spec.ts`: `commentsIds.xml` emitted, content-type and relationship registered

Full suite passes (`npx vitest run`): 196 files, 1032 tests. `npm run build`, lint, and prettier are clean.

## Notes

- There is no behavior change for documents that do not set `durableId`. The part is only created when at least one comment opts in.
- `durableId` is a string so callers can supply Word-style 8-character hex ids (for example `"12AB34CD"`) verbatim, without numeric coercion.

## Summary by CodeRabbit

* **New Features**
  * Added support for stable, persistent comment identifiers in DOCX output.
  * When available, the export now generates and wires the `word/commentsIds.xml` part so comment IDs remain consistent across edits, including for threaded comments.
* **Tests**
  * Expanded test coverage to validate creation/omission behavior for `commentsIds.xml`, durable-id mapping, and related XML parts and relationships.

### DIFF
--- a/src/export/packer/next-compiler.spec.ts
+++ b/src/export/packer/next-compiler.spec.ts
@@ -218,6 +218,30 @@ describe("Compiler", () => {
             expect(commentsIdsText).to.contain('w16cid:durableId="33334444"');
         });
 
+        it("should emit commentsIds.xml falling back to paraId for comments without a durableId", { timeout: 99999999 }, async () => {
+            const file = new File({
+                sections: [],
+                comments: {
+                    children: [
+                        { id: 0, children: [new Paragraph("with durable")], durableId: "12AB34CD" },
+                        { id: 1, children: [new Paragraph("without durable")] },
+                    ],
+                },
+            });
+            const zipFile = compiler.compile(file);
+            const fileNames = Object.keys(zipFile.files).map((f) => zipFile.files[f].name);
+
+            expect(fileNames).to.include("word/commentsIds.xml");
+
+            const commentsIdsText = await zipFile.file("word/commentsIds.xml")?.async("text");
+            // Comment WITH a durableId keeps its durableId (paraId 00000001 for id 0)
+            expect(commentsIdsText).to.contain('w16cid:paraId="00000001"');
+            expect(commentsIdsText).to.contain('w16cid:durableId="12AB34CD"');
+            // Comment WITHOUT a durableId falls back to its generated paraId (00000002 for id 1)
+            expect(commentsIdsText).to.contain('w16cid:paraId="00000002"');
+            expect(commentsIdsText).to.contain('w16cid:durableId="00000002"');
+        });
+
         it("should call the format method X times equalling X files to be formatted", () => {
             // This test is required because before, there was a case where Document was formatted twice, which was inefficient
             // This also caused issues such as running prepForXml multiple times as format() was ran multiple times.
--- a/src/export/packer/next-compiler.spec.ts
+++ b/src/export/packer/next-compiler.spec.ts
@@ -153,6 +153,71 @@ describe("Compiler", () => {
             expect(commentsExtendedText).to.contain("w15:paraIdParent");
         });
 
+        it("should not include commentsIds.xml when comments have no durableId", () => {
+            const file = new File({
+                sections: [],
+                comments: {
+                    children: [{ id: 0, children: [new Paragraph("comment")] }],
+                },
+            });
+            const zipFile = compiler.compile(file);
+            const fileNames = Object.keys(zipFile.files).map((f) => zipFile.files[f].name);
+
+            expect(fileNames).to.not.include("word/commentsIds.xml");
+        });
+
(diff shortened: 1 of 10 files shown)


alexvcasillas:
Addressed the nitpicks in c74aa32:

- **Assert XML shape in `file.spec.ts`**: the positive `durableId` test now formats `CommentsIds` and asserts the actual tree (`w16cid:commentsIds` → `w16cid:commentId` with `w16cid:paraId` / `w16cid:durableId`), instead of only checking presence.
- **Mixed durable / non-durable coverage in `next-compiler.spec.ts`**: added a test with one comment carrying a `durableId` and one omitting it, asserting the omitted comment falls back to its generated `paraId` in `word/commentsIds.xml`.
- **Path alias in `file.ts`**: left as a relative import. The `@file/` alias is used for reaching across top-level areas, but imports within `src/file/` use relative paths (every other import in `file.ts`, and sibling files like `header-wrapper.ts`, do the same). The flagged line sits next to the existing relative `./paragraph/run/comment-run` import and matches that local convention, so switching only this one to an alias would be inconsistent.

Full suite passes (`npx vitest run`): 196 files, 1033 tests. Build, ESLint, and Prettier are clean.

alexvcasillas:
Hi @dolanmiu, a small ask on releases rather than on this change itself.

This merged on 7 August, and the latest published version is still `9.7.1` from 27 May, so `commentsIds.xml` is not yet reachable from npm. We are carrying it as a local patch in the meantime, and we would rather depend on a release than keep patching.

Is a release planned? Happy to help with whatever would make one easier, and no rush if the timing is not right.

For context, we have one more bookmark id fix open in #3502. Together those are the last two things keeping us on a patched copy of `docx` instead of the published package.

Issue #3015, opened 2025-03-26
Table columns width are not working in case of when I opened docx file on mac and google docs


ankushcloudanalogy:
const createPaginatedTables = (
  headers: string[],
  rows: any[][],
  colors: typeof TEMPLATE_COLORS.crux,
  pageTitle: string,
  tableIndex: number,
): {tables: Table[]; continuationTitles: Paragraph[]} => {
  const tables: Table[] = [];
  const continuationTitles: Paragraph[] = [];

  // Page width calculations
  const pageWidth = convertInchesToTwip(8.5);
  const margins = convertInchesToTwip(1);
  const availableWidth = pageWidth - margins * 2;

  // Set equal column width
  const columnWidth = Math.floor(availableWidth / headers.length);

  for (let i = 0; i < rows.length; i += MAX_RECORDS_PER_PAGE) {
    const chunk = rows.slice(i, i + MAX_RECORDS_PER_PAGE);

    if (i > 0) {
      continuationTitles.push(
        new Paragraph({
          text: `${pageTitle} - Table ${tableIndex + 1} (Continued)`,
          heading: HeadingLevel.HEADING_2,
          style: 'Heading2',
          spacing: {before: 400, after: 200},
        }),
      );
    }

    const tableRows: TableRow[] = [];

    // Header Row
    tableRows.push(
      new TableRow({
        tableHeader: true,
        cantSplit: true,
        height: {value: 500, rule: HeightRule.EXACT},
        children: headers.map(
          (header) =>
            new TableCell({
              children: [
                new Paragraph({
                  text: header,
                  style: 'TableHeader',
                  alignment: AlignmentType.CENTER,
                  spacing: {before: 100, after: 100},
                }),
              ],
              shading: {
                fill: `#${colors.tableHeader}`,
                type: ShadingType.CLEAR,
              },
              width: {size: columnWidth, type: WidthType.DXA},
            }),
        ),
      }),
    );

    // Data Rows
    chunk.forEach((row, rowIndex) => {
      tableRows.push(
        new TableRow({
          cantSplit: true,
          height: {value: 400, rule: HeightRule.AUTO},
          children: row.map(
            (cell) =>
              new TableCell({
                children: [
                  new Paragraph({
                    text: cell?.toString() || 'N/A',
                    style: 'TableCell',
                    alignment: AlignmentType.LEFT,
                    spacing: {before: 50, after: 50},
                  }),
                ],
                shading: {
                  fill:
[...]

Issue #349, opened 2019-07-07
table width property not working for google and microsoft doc


liquidslr:
Hi, all I have trying around to create a table with width 100% having two columns with width 50% each, but the table is not taking full doc width in Microsoft doc and google doc. Here is the code that i am using 

````
let rowmatch = Object.keys(state.match_keys).length;
  const tablematch = doc.createTable(rowmatch, 2).setWidth('pct', '100%');

  Object.keys(state.match_keys).map((key, i) => {
    const parakey = new docx.Paragraph().spacing({
      before: 20,
      after: 20
    });

    parakey.addRun(
      new docx.TextRun(`${key}`)
        .font("Verdana")
        .size(20)
        .color("#2b2b2b")
    );

    parakey.addRun(new docx.TextRun("").break());
    const paraval = new docx.Paragraph().spacing({
      before: 20,
      after: 20
    });

    paraval.addRun(
      new docx.TextRun(`${state.match_keys[key]}`)
        .font("Verdana")
        .size(20)
        .color("#2b2b2b")
    );

    paraval.addRun(new docx.TextRun("").break());

    tablematch.getCell(i, 0).addContent(parakey).CellProperties.setWidth('50%', 'pct');
    tablematch.getCell(i, 1).addContent(paraval).CellProperties.setWidth('50%', 'pct');
  });
````

I am using  docx  version  4.7.1
I have attached the screenshot for reference.

![Screenshot 2019-07-07 at 2 18 32 PM](https://user-images.githubusercontent.com/26876811/60766096-2cf71280-a0c2-11e9-8a14-92685cd0dafc.png)

**Comment by carlbolduc** [comment]
Tables are only supported in Word. I use Tab Stops to work around this, they work perfectly in both Pages and Google Docs.

**Comment by faradaytrs** [comment]
@carlbolduc Is the any way to make it supported. And hwo do you use Tab Stops to work around that?

**Comment by dolanmiu** [comment]
Table on google docs aren't supported at this moment

Would appreciate it if anyone can figure out why it has this weird behaviour

**Comment by fpirsch** [comment]
@dolanmiu it looks like it's a unit problem, Google Docs does not like percents.
MS Word generates xml like this (which is GDocs compatible) :
```
<w:tblW w:w="0" w:type="auto"/>
...
<w:gridCol w:w="1795"/>
```
And docx generates this :
```
<w:tblW w:type="pct" w:w="80%"/>
...
<w:gridCol w:w="100"/>
```

[...]

Issue #216, opened 2018-12-07
Wrong Table width when preview file & uploading Google Docs


huynhtansi:
Hi, dolanmiu. Thank you for awesome library.

I has a problem when creating a full-width table with code:

`
const table = doc.createTable(3, headers.length)
table.setWidth(WidthType.PERCENTAGE, '100%')
`

It works fine when open in Microsoft Word, but wrong width when preview in Google Docs or Finder (I'm using MacOS).

Microsoft Word: 

<img width="799" alt="screen shot 2018-12-07 at 5 12 03 pm" src="https://user-images.githubusercontent.com/16728346/49641656-6344b480-fa43-11e8-85c2-1e282c30f437.png">

Google Docs:

<img width="816" alt="screen shot 2018-12-07 at 4 58 57 pm" src="https://user-images.githubusercontent.com/16728346/49641670-735c9400-fa43-11e8-88c2-5a9c04618611.png">

Finder Preview:

<img width="806" alt="screen shot 2018-12-07 at 5 12 29 pm" src="https://user-images.githubusercontent.com/16728346/49641683-7a83a200-fa43-11e8-8b97-dabf9836887f.png">

dolanmiu:
Yes, Gogole docs, or finder is not supported

Only Microsoft Word so far

If you can figure out how to get the features working in each, that would be amazing

carlbolduc:
I use Tab Stops to work around this, they work perfectly in both Pages and Google Docs.

vahid18u:
Would you please write somewhere in the documentation with **bold** font that this package does not support google docs ? It took more than 2 days for me to find it out and see this issue with table width. Thanks

WilliamPeralta:
@vahid18u if you want you can add the warning on the documentation yourself and open a PR,

baloo887:
> I use Tab Stops to work around this, they work perfectly in both Pages and Google Docs.

Hi @carlbolduc , can you please explain how you use tab stops in your workaround? Everything else is working fine in my document, but tables. If your workaround works, you made my day!!

carlbolduc:
> Hi @carlbolduc , can you please explain how you use tab stops in your workaround? Everything else is working fine in my document, but tables. If your workaround works, you made my day!!

Here is an example, I first create a line like this:
```
const line = this.doc.createParagraph().bullet().maxRightTabStop();
```
I can then add content to it:
```
line.createTextRun('Stuff...');
```
And use tab stops to simulate a table:
```
[...]

PR #3476, opened 2026-07-22
fix(table): emit numeric pct widths in fiftieths of a percent

### LINKED ISSUE #1457: WidthType.PERCENTAGE doesn't work for tables in Google Docs or Apple Pages
When creating the following table using percent widths it works fine in Word but doesn't in Google Docs or Apple Pages:

```js
const fs = require('fs');
const {execSync} = require('child_process');
const {
	Document,
	Packer,
	Paragraph,
	Table,
	TableCell,
	TableRow,
	WidthType
} = require('docx');

const dxaWidthTable = new Table({
	width: {
		size: 100,
		type: WidthType.PERCENTAGE,
	},
	columnWidths: [3505, 5505],
	rows: [
		new TableRow({
			children: [
				new TableCell({
					width: {
						size: 3505,
						type: WidthType.DXA,
					},
					children: [new Paragraph('Hello')],
				}),
				new TableCell({
					width: {
						size: 5505,
						type: WidthType.DXA,
					},
					children: [],
				}),
			],
		}),
		new TableRow({
			children: [
				new TableCell({
					width: {
						size: 3505,
						type: WidthType.DXA,
					},
					children: [],
				}),
				new TableCell({
					width: {
						size: 5505,
						type: WidthType.DXA,
					},
					children: [new Paragraph('World')],
				}),
			],
		}),
	],
});

const percentWidthTable = new Table({
	width: {
		size: 100,
		type: WidthType.PERCENTAGE,
	},
	columnWidths: [90, 10],
	rows: [
		new TableRow({
			children: [
				new TableCell({
					width: {
						size: 90,
						type: WidthType.PERCENTAGE,
					},
					children: [new Paragraph('Hello')],
				}),
				new TableCell({
					width: {
						size: 10,
						type: WidthType.PERCENTAGE,
					},
					children: [],
				}),
			],
		}),
		new TableRow({
			children: [
				new TableCell({
					width: {
						size: 90,
						type: WidthType.PERCENTAGE,
					},
					children: [],
				}),
				new TableCell({
					width: {
						size: 10,
						type: WidthType.PERCENTAGE,
					},
					children: [new Paragraph('World')],
				}),
			],
		}),
	],
});


const doc = new Document({
	sections: [
		{
			children: [
				new Paragraph({ text: 'Table with DXA widths' }),
				dxaWidthTable,
				new Paragraph({ text: 'Table with percentage widths' }),
				percentWidthTable,
			],
		},
	],
});


const fileName = `/tmp/test-${(new Date().getTime()).toString(36)}.docx`;

(async () => {

	const buffer = await Packer.toBuffer(doc);

	fs.writeFileSync(fileName, buffer);
	execSync(`open ${fileName}`);
	execSync(`open -a /Applications/Pages.app ${fileName}`);

})();
```

The result looks like this:

Screen Shot 2022-03-17 at 1 56 49 PM

Node.js: 14.17.5
docx: v 7.3.0

omartuhintvs:
## What

`createTableWidthElement` serialises a numeric `WidthType.PERCENTAGE` size as a percent **string** (`w:w="50%"`). This changes it to emit the value in **fiftieths of a percent** (`50% → 2500`), which is the form Word itself writes.

Fixes #1457 (also the root cause behind #349, #216, #3015).

## Why

For `w:tblW` / `w:tcW`, `@w:w` is `ST_MeasurementOrPercent`, whose union permits *both* an `ST_Percentage` string (`"50%"`) and an `ST_UnqualifiedPercentage` integer in fiftieths. So the current `"50%"` output is schema-valid — this is an **interoperability** issue, not a spec violation:

- **Word** writes `pct` widths as the fiftieths integer.
- **Stricter consumers** — Google Docs' `.docx` importer and Apple's shared OOXML parser (Pages, QuickLook, Mail preview, iCloud) — misread the `"50%"` string, fall back to `tblGrid`, and collapse every column to ~1 character wide (text stacked vertically). This is the long-standing "tables broken in Google Docs" report.

Emitting the integer matches Word and renders correctly everywhere, with no visual change in Word/LibreOffice.

## Compatibility

- The rendered percentage is unchanged (`{ size: 50, type: PERCENTAGE }` still means 50%).
- Only the numeric branch changes. Callers who deliberately want the literal string can still pass a `Percentage` (e.g. `{ size: "50%", type: PERCENTAGE }`) — that path is untouched.
- Note: this does change the emitted XML bytes for existing numeric-percentage callers (`"100%"` → `5000`). Two existing specs asserted the old string output and are updated. I recognise `%` was introduced deliberately in #282; happy to gate this behind an option instead if you'd prefer to preserve the exact previous bytes by default.

## Tests

- New `src/file/table/table-width.spec.ts` covering: numeric pct → fiftieths, fractional rounding, the `Percentage`-string path still emitting the string, and `dxa` untouched.
- Updated the two specs that asserted the old string (`table.spec.ts`, `table-cell-margin.spec.ts`).
- Full suite green: **1023 passed / 197 files**. Lint clean on changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved table and cell percentage-width formatting for better compatibility with Word and other document editors.
  * Numeric percentage widths now use the correct Word-compatible representation, including proper rounding.
  * Preserved explicitly provided percentage strings and DXA width values.
  * Updated coverage to verify common and edge-case table width scenarios.

### DIFF
--- a/src/file/table/table-properties/table-cell-margin.spec.ts
+++ b/src/file/table/table-properties/table-cell-margin.spec.ts
@@ -135,7 +135,7 @@ describe("TableCellMargin", () => {
 
             const tree = new Formatter().format(cellMargin!);
             expect(tree).to.deep.equal({
-                "w:tcMar": [{ "w:left": { _attr: { "w:type": "pct", "w:w": "5%" } } }],
+                "w:tcMar": [{ "w:left": { _attr: { "w:type": "pct", "w:w": 250 } } }],
             });
         });
     });
--- a/src/file/table/table-width.spec.ts
+++ b/src/file/table/table-width.spec.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+
+import { Formatter } from "@export/formatter";
+
+import { WidthType, createTableWidthElement } from "./table-width";
+
+describe("createTableWidthElement", () => {
+    it("writes a numeric pct width in fiftieths of a percent, matching Word (#1457)", () => {
+        // A "50%" string is misread by Google Docs / Apple Pages and collapses
+        // the table; Word writes 50% as the integer 2500.
+        const tree = new Formatter().format(createTableWidthElement("w:tblW", { size: 50, type: WidthType.PERCENTAGE }));
+
+        expect(tree).to.deep.equal({ "w:tblW": { _attr: { "w:type": "pct", "w:w": 2500 } } });
+    });
+
+    it("rounds fractional percentages to the nearest fiftieth", () => {
+        const tree = new Formatter().format(createTableWidthElement("w:tblW", { size: 33.3, type: WidthType.PERCENTAGE }));
+
+        expect(tree).to.deep.equal({ "w:tblW": { _attr: { "w:type": "pct", "w:w": 1665 } } });
+    });
+
+    it("still emits the literal percent string when the caller passes a Percentage", () => {
+        const tree = new Formatter().format(createTableWidthElement("w:tblW", { size: "50%", type: WidthType.PERCENTAGE }));
+
+        expect(tree).to.deep.equal({ "w:tblW": { _attr: { "w:type": "pct", "w:w": "50%" } } });
+    });
+
+    it("leaves dxa widths untouched", () => {
+        const tree = new Formatter().format(createTableWidthElement("w:tblW", { size: 5000, type: WidthType.DXA }));
+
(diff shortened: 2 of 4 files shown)


dolanmiu:
Awesome. Did not know this quirk with Word/Google Docs

Good stuff. Can we make it non-draft?

omartuhintvs:
it did not fix the issue :(

